PR #3, opened 2022-03-21
Set name in `config/application.rb` via initial setup

gazayas:
・Addresses the TODO in `config/application.rb`

### Details
I went with the naming convention already used in `bin/configure` and changed `bullettrain` to `untitled_application`.

As a side note, I have also addressed `sanitized_allowed_protocols` [here](https://github.com/bullet-train-co/bullet-train-tailwind-css/pull/580) in the second part of the details.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -123,6 +123,7 @@ kebab_case = variable.tr("_", "-")
 puts ""
 puts green "Replacing instances of \"Untitled Application\" with \"#{human}\" throughout the codebase."
 replace_in_file("./.circleci/config.yml", "untitled_application", variable)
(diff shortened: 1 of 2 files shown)


andrewculver:
@gazayas OK, this will work, but we need to get rid of any potential `_`, so maybe do a `.gsub("_", "")` or something. See https://stackoverflow.com/questions/9142309/may-url-scheme-contain-an-underscore for context.

gazayas:
@andrewculver We also have `kebab_case` which also appears to be a valid URL scheme according to the Stack Overflow post, but I went ahead with the `gsub("_", "")` for now.

PR #20, opened 2022-04-04
Install foreman in `bin/setup`

gazayas:
・Addresses TODO in Gemfile

### Details
Rails installs foreman manually when installing esbuild.
There are some commands that are run in the main Rails repo, but since esbuild is optional, the logic for installing foreman was actually in [jsbundling-rails](https://github.com/rails/jsbundling-rails/blob/b6ec0980ff573203ac1031074e57bc59be5acbf5/lib/install/install.rb#L34):
```ruby
if Rails.root.join("Procfile.dev").exist?
  append_to_file "Procfile.dev", "js: yarn build --watch\n"
else
  say "Add default Procfile.dev"
  copy_file "#{__dir__}/Procfile.dev", "Procfile.dev"

  say "Ensure foreman is installed"
  run "gem install foreman"
end
```

Bullet Train already ships with `Procfile.dev` configured with everything we need, so I just added the installation command in `bin/setup`, since it's basically the same thing `js-bundling` is doing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -67,9 +67,6 @@ group :development do
 
   # Speed up commands on slow machines / big apps [https://github.com/rails/spring]
   # gem "spring"
-
-  # TODO Not sure why we need to add this when it's not in the Gemfile of a new Rails application.
-  gem "foreman"
 end
 
 group :test do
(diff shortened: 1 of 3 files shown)


andrewculver:
@gazayas Thanks for all the context on this. I'm actually going to leave it in the Gemfile now that I understand the context.

andrewculver:
See https://github.com/bullet-train-co/bullet_train/commit/5f59e876a5981cb01cda0ab45668ad93d0375b56 .

PR #21, opened 2022-04-04
Remove `collection_actions` from routes file (WIP)

gazayas:
・Addresses the TODO in `config/routes.rb` about `collection_actions`

### Details
Super Scaffolding already has this in its routes, but if we scaffold `collection_actions` after we initialize the app, I'm assuming the routes hook will be generated in the routes file here and not in `bullet_train-super_scaffolding`. I ended up making this helper so we could remove more code from the file to not overwhelm any developers, but if you think there's a better way to take care of the issue, just let me know.

### WIP
If decide to use the helper, we'll have to edit how `collection_actions` is super scaffolded as well, so I can work on that once we decide what direction to go in.


jagthedrummer:
@gazayas I don't fully understand why that variable exists in the first place. Is it that super scaffolding will generate code that literally refers to a magic variable (instead of populating an `only` option directly)?

Like, does SS sometimes do this:
```
resources :widgets, only: collection_actions
```

Instead of doing:
```
resources :widgets, only: [:index, :new, :create]
```

gazayas:
> Is it that super scaffolding will generate code that literally refers to a magic variable (instead of populating an only option directly)?

@jagthedrummer Correct, here is an example where it is super scaffolded:

https://github.com/bullet-train-co/bullet_train-core/blob/4e2ef32dbfd3e9597b4904bd340c731412a1dc52/bullet_train-super_scaffolding/lib/scaffolding/routes_file_manipulator.rb#L340

gazayas:
Closing in favor of #834.

PR #24, opened 2022-04-04
Enable native view annotation. Remove Xray.

andrewculver:


ParamagicDev:
🥳 🎉 🎈 

https://blog.saeloun.com/2020/05/11/rails-support-annotates-html-output-with-template-file-names.html#with-rails-61

^ For the curious.

Also of note, last time i checked it only works with ERB, not sure what haml / slim support looks like.

andrewculver:
@ParamagicDev Thankfully for us we don't have to worry about those. Neither are supported in BT. :-)

PR #27, opened 2022-04-06
Add React

### LINKED ISSUE #26: Add React
This came up in the Discord chat, so I thought I'd open an issue here.

gazayas:
・Closes #26.

### Details
Packages added:
1. `react`
2. `react-dom`
3. `remount`

There are different approaches to adding React (like pinning libraries with [importmaps](https://github.com/rails/importmap-rails#how-do-importmaps-work)), but I went with this one since we're using esbuild. Since we don't have a specific use-case where we want to apply it right now I haven't created any components or written tests. If we want to go with this approach though, I can add those if needed.

Link to [remount](https://www.npmjs.com/package/remount).

### Example
You can add the following to try a component.
`app/javascript/react/components/foo.jsx`:
```javascript
import * as React from 'react'
import * as ReactDOM from 'react-dom/client'

const Foo = () => {
  return (<div>A new React component here</div>)
}

export default Foo
```

`app/javascript/react/index.js`:
```javascript
import { define } from 'remount'
import Foo from "./components/foo"

define({ 'foo-component': Foo })
```

`app/views/account/teams/show.html.erb`:
```html
<!-- Add this line at the top of the file -->
<foo-component />
```

Result:
Screenshot from 2022-04-06 09-10-08

### React 18 error
Since React 18 came out recently, some of the syntax has changed:
Screenshot from 2022-04-06 09-26-32

Here's the [link they provided](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis). We're not directly using `ReactDOM.render` here though, so we'll have to find a workaround if we want to use 18 and esbuild.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -9,6 +9,9 @@ import "@hotwired/turbo-rails"
 import "./controllers"
 import "./channels"
 
+// React Components
+import "./react/index.js"
+
 Rails.start()
 ActiveStorage.start()
 
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
     "postcss-extend-rule": "^4.0.0",
     "postcss-import": "^14.0.2",
     "postcss-nested": "^5.0.6",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "remount": "^0.11.0",
     "stimulus-reveal": "^1.2.12",
     "tailwindcss": "^3.0.23"
   },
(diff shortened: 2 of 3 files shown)


BilalBudhani:
Why not just create a wiki or add a guide in the documentation on how to add React to the project instead of introducing directly in the project?

andrewculver:
@BilalBudhani Yup, agree. Added my thoughts at https://github.com/bullet-train-co/bullet_train/issues/26 .

PR #28, opened 2022-04-06
Add open-gem script

gazayas:
Although we probably don't want to edit these files directly, I think it would be helpful if we had a quick command to access these files to make things easier to navigate.


andrewculver:
@gazayas Thanks. Great idea.

PR #66, opened 2022-04-17
Added missing loader for woff2.

nathanclark:
This is needed after installing FontAwesome Pro. Otherwise the esbuild fails.

See the following screenshot for the encountered error:
image


andrewculver:
@nathanclark Thanks for this!

Issue #16, opened 2022-03-31
ActionView::Template::Error (The asset "application.light.css" is not present in the asset pipeline.


JackWetson:
First run of bin/dev after bin/configure and bin/setup

`13:42:22 web.1       | ActionView::Template::Error (The asset "application.light.css" is not present in the asset pipeline.`

andrewculver:
@JackWetson I think this is a race condition. The light theme takes a little longer to compile, because it has to run `rake bt:link` first. Can you try refreshing after giving it a little bit to finish compiling and see if it appears?

JackWetson:
> @JackWetson I think this is a race condition. The light theme takes a little longer to compile, because it has to run `rake bt:link` first. Can you try refreshing after giving it a little bit to finish compiling and see if it appears?

@andrewculver can confirm it's race condition indeed!

andyw8:
I also ran into this, refreshing resolved it.

andrewculver:
@gazayas I'm not really sure what we could do to fix this. The web server comes up before the stylesheets are done compiling. 🤷‍♂️ I'll let you noodle on this one a bit.

andyw8:
Perhaps `bundle exec rake bt:link` should be moved to `bin/dev` or `bin/configure`?

PR #72, opened 2022-04-18
Move bt:link to start of bin/dev

gazayas:
Address race condition in #16.

Although I don't think this entirely takes care of the race condition, I think it's better than having it in `Procfile.dev`.

I confirmed that even though `bt:link` finished before the server was fired up, the error still showed up on the page when trying to access `localhost:3000` as soon as I could. However, if we have `bt:link` in `Procfile.dev`, it'll cut down on the time the stylesheet build happens once foreman starts, so I think this is a better option than having it paired with the css build command.


andrewculver:
@gazayas I accidentally did this myself. 🤦‍♂️

gazayas:
@andrewculver Oh no! I saw the commit :sweat_smile:
Now that we have overmind, I shouldn't have to worry about this, right?

PR #82, opened 2022-04-20
Makes Overmind default

### LINKED ISSUE #79: Default to Overmind.
We should encourage folks to install Overmind as part of `bin/setup`, allow them to skip it and fallback to Foreman, but if Overmind is present we should use it.

BilalBudhani:
Fixes #79


andrewculver:
@BilalBudhani Thanks so much for this! 🙏

PR #126, opened 2022-05-05
Use the free version of Rails Autoscale

adamlogic:
When deploying to Heroku for the first time, there's no need to start with a paid Rails Autoscale plan. The free plan supports 20 autoscale events per month at any scale, and will prompt if there is a need to upgrade.


andrewculver:
@adamlogic Thanks for this!

PR #147, opened 2022-05-13
Test PR for checking db schema git diff

gazayas:
I plan on closing this after opening the actual PR for the fix, but CircleCI isn't running on my forked repo so I'm just opening here so we can check.


andrewculver:
@gazayas Very nice!

PR #148, opened 2022-05-13
Update the CircleCI `db/schema.rb` check to output the diff on fail

gazayas:
This is the fix by itself for outputting `db/schema.rb`'s git diff.


andrewculver:
@gazayas This is awesome.

gazayas:
@andrewculver Thanks :raised_hands:

PR #169, opened 2022-05-19
Switch system tests to Cuprite

existentialmutt:
This PR switches the backend for system tests to use [Cuprite](https://github.com/rubycdp/cuprite) as a more reliable alternative to Selenium.  It is generally drop-in.  There are a few differences in the driver API that required some updates to the test code:

- Accessing browser console logs (for `assert_no_js_errors`) requires a custom logger (based on https://github.com/rubycdp/cuprite/issues/113#issuecomment-753598305)
- Cuprite doesn't support `page.driver.browser.switch_to.alert.accept`, but we can use the generic Capybara `accept_alert` instead
- I updated the code for typing in select2 elements, moving from send_keys to `page.driver.browser.keyboard.type`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,7 @@ jobs:
     parallelism: 16
     steps:
       - browser-tools/install-chrome
+      - browser-tools/install-chromedriver
       - checkout
       - restore_cache: *restore_bundler_cache
       - restore_cache: *restore_yarn_cache
@@ -139,6 +140,7 @@ jobs:
     parallelism: 7
     steps:
       - browser-tools/install-chrome
+      - browser-tools/install-chromedriver
       - checkout
 
       # Restore dependency caches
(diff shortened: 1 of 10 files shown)


andrewculver:
Encouraging discuss here: https://twitter.com/andrewculver/status/1528026237976776704?s=21&t=jxSRIVJ60k9L_txuPSbYbA

andrewculver:
@existentialmutt Based on the speed improvements *and* the issues people have reported to me, my gut tells me that we need to make this default but optional. Here's what I think we should do:

1. Make the test suite pass on either Selenium _or_ Cuprite.
2. Make the test suite configuration pick up on whether the app has `gem "cuprite"` or `gem "webdrivers"` and `gem "selenium-webdriver"` and use whichever one is included without any changes to `test/application_system_test_case.rb`
3. Make Cuprite the default, but leave the Selenium gems in the `Gemfile` commented out with a note that we use Cuprite because of its speed, but some people have trouble with it so folks can fallback to Selenium if they need to.
4. Add a new CI step to `.circleci/config.yml` that only runs when `BT=true` in the environment and make that step comment out Cuprite and uncomment Selenium in the `Gemfile` and then run the test suite as normal. Basically just testing that if people try to fallback to Selenium it'll still work for them.

I'm going to set `BT=true` on our CI environment now so you can count on that being there when you push up an update to this PR.

existentialmutt:
Sounds good @andrewculver, I like the idea of leaving Selenium in as a fallback.

Pretty sure we can put the gems back in the gemfile and then just switch the Capybara driver.  Could do it based on an ENV var defaulting to cuprite

I’ll get going on that setting that up next week

existentialmutt:
Thinking about it more your solution of checking on installed gems is pretty slick.  I’ll go ahead with that

andrewculver:
@existentialmutt I've made some major improvements in `main` with the existing test suite. Can you also rebase this the next time you work on it?

existentialmutt:
@andrewculver This should be good to go.  Selenium is the default and it will switch to Cuprite if you switch the gems in the Gemfile.

PR #173, opened 2022-05-21
swap membership/team locale keys to correct files

wesgarrison:


andrewculver:
@wesgarrison 🤦🏻‍♂️, but I love that you sent this… you must be up to something. :-)

Issue #176, opened 2022-05-21
Upgrade to a next-gen Docker convenience image on CircleCI.


andrewculver:
<img width="820" alt="Screen Shot 2022-05-21 at 9 07 13 AM" src="https://user-images.githubusercontent.com/217218/169660016-39938a75-7fb6-406a-9eaf-63c7b7a93b74.png">

PR #180, opened 2022-05-23
Update CircleCI images

gazayas:
Labeling this as a WIP to check CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.1
 orbs:
-  ruby: circleci/ruby@0.1.2
+  ruby: cimg/ruby@0.1.2
   browser-tools: circleci/browser-tools@1.1
 aliases:
   - &restore_bundler_cache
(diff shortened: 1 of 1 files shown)


gazayas:
Closes #176.

## Docs
[The cimg docs for PostgreSQL](https://circleci.com/developer/images/image/cimg/postgres)
[The cimg docs for Redis](https://circleci.com/developer/images/image/cimg/redis)

## Details
There wasn't much that needed to happen here, we just needed to change the prefix and specify the version for the images we're using.

The `ruby` orb (which isn't a Docker image) still uses the `circleci` prefix, so this doesn't need to be changed.

Also, there were different options for using `browser-tools`, but it has been condensed to one option. We already use that line, and we also have the `browser-tools` orb it says to include, so we don't need to change anything for the `browser-tools` either.

The link provided in the issue also had a link to [this CircleCI blog post](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/) about next-gen convenience images, so it's worth a read if we want to go in more depth with it.

PR #186, opened 2022-05-24
Add support for running the test suite on GitHub actions.

andrewculver:
I'm really curious to see how fast we can get the test suite running on GitHub actions compared to CircleCI. This PR is a rebase of @MikeRogers0's PR on the original Bullet Train repository at https://github.com/bullet-train-co/bullet-train/pull/439. There are some comments from Mike on that PR.

Miss you, Mike! ❤️

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,12 @@ jobs:
     name: Rails Tests
     runs-on: ubuntu-20.04
 
+    strategy:
+      fail-fast: false
+      matrix:
+        ci_node_total: [16]
+        ci_node_index: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+
     services:
       # Setup a PostgreSQL data for us to test against
       postgres:
@@ -70,8 +76,13 @@ jobs:
           yarn run light:build:css
 
       # Run the tests
-      - name: Running The Tests
+      - name: Minitest
         env:
           RAILS_ENV: test
           DATABASE_URL: postgres://postgres:postgres@127.0.0.1:5432/
-        run: bundle exec rails test
+          KNAPSACK_PRO_TEST_SUITE_TOKEN_MINITEST: ${{ secrets.KNAPSACK_PRO_TEST_SUITE_TOKEN_MINITEST }}
+          KNAPSACK_PRO_CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
+          KNAPSACK_PRO_CI_NODE_INDEX: ${{ matrix.ci_node_index }}
+          KNAPSACK_PRO_FIXED_QUEUE_SPLIT: true
+          KNAPSACK_PRO_LOG_LEVEL: info
+        run: bundle exec rake knapsack_pro:queue:minitest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,6 +93,5 @@ jobs:
 
       - name: Minitest for Super Scaffolding
         run: |
-          bundle add spring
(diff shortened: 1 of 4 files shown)


andrewculver:
@gazayas Let me know if you need to be made an admin on the organization in order to do stuff with GitHub actions. I haven't played around with it at all. Knapsack Pro has an article about parallelism on GitHub actions at https://docs.knapsackpro.com/2021/run-minitest-on-github-actions-with-parallel-jobs-using-build-matrix .

gazayas:
@andrewculver Got the tests to pass, I'll see if I can make them parallel with the link you provided.

gazayas:
The Super Scaffolding Minitest fails in GitHub Actions, but it's also failing locally for me. I tried running the Super Scaffolding system tests on the main branch, and I got the same errors:
```
RuntimeError: Wrapped undumpable exception for: ActionView::Template::Error: undefined method `new_account_team_personality_disposition_path' for #<ActionView::Base:0x0000000003a930>

              recipient.public_send(method, *args, options)
                       ^^^^^^^^^^^^
Did you mean?  new_account_personality_disposition_note_path
               new_account_insight_personality_disposition_path
               new_account_personality_disposition_note_url
               new_account_insight_personality_disposition_url
```

Besides this, the GitHub Actions tests are passing, although I'm not sure why the tests are passing in CircleCI.

Something else to note is that trying to run the tests in separate commands gave me a database connection issue. I solved this by lumping both the normal Minitest tests and the Super Scaffolding Minitest tests together. Here's the error I was getting after the Super Scaffolding setup script finished:
```
1. If you would like the table view you've just generated to reactively update when a Test File is updated on the server, please edit `app/models/team.rb`, locate the `has_many :test_files`, and add `enable_updates: true` to it.

DEBUGGER: Attaching after process 3347 fork to child process 3365
Running via Spring preloader in process 3365
rake aborted!
ActiveRecord::ConnectionNotEstablished: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
```

andrewculver:
@gazayas I think the Super Scaffolding test failure is actually a legit failure we need to fix. Opened an issue for it: https://github.com/bullet-train-co/bullet_train/issues/192

gazayas:
@andrewculver Got it, I'll look over the test again.

gazayas:
@andrewculver Since we're getting an SSH error, I think we'll have to generate a private key for the workflow:
```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```
In the last commit, I added the call to the environment variable in the workflow for use with [webfactory/ssh-agent](https://github.com/marketplace/actions/webfactory-ssh-agent), so we'll just need to add the environment variable according to the docs.

[Install SSH Key](https://github.com/marketplace/actions/install-ssh-key) is another option in case we want to use something else.

kaspth:
I haven't looked too closely at this yet, but it looks like there's an official `setup-node` action, which has yarn support https://github.com/marketplace/actions/setup-node-js-environment

jagthedrummer:
@andrewculver @gazayas @kaspth Is this still something we want to do? If so, I don't mind either picking up this PR or taking a fresh stab at it. (I already have GitHub actions running CI in my own BT project, so hopefully I could just plop that config in and it will mostly work.)

gazayas:
I'm all for this being implemented later. We've been making some progress with the current CI setup, so it might be best to wait and decide on a good workflow there and then apply a similar test setup in GitHub actions after we feel confident with it.

jagthedrummer:
Considering that this PR is more than a year old, and in the interests of reducing in-flight WIP, I'm going to go ahead and close this one. We have issue #764 so we don't lose track of the idea, and it points to this PR if there's anything here we want to retrieve later.

PR #189, opened 2022-05-24
Build assets in `bin/configure`

gazayas:
When initially making a Bullet Train application, developers run the following commands:
`bin/configure`
`bin/setup`

However, unit tests will fail if they haven't run `bin/dev` yet because the assets haven't been compiled yet.

This step compiles the assets in the configure step so developers don't get confused in case they try to run tests before ever starting the server, assuming that things are broken.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -177,6 +177,9 @@ end
 puts green "Running `yarn install`."
 stream "yarn install"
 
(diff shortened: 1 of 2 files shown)


gazayas:
@andrewculver Done!

Issue #222, opened 2022-06-12
running bin/configure results in error: cannot load colorize


esmale:
Cloned the starter repo from scratch and ran `bin/configure` as the starter documentation instructs.  Wind up with this error:

```
rescueready@RescueReady rescue_ready % bin/configure
<internal:/Users/rescueready/.rbenv/versions/3.1.2/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- colorize (LoadError)
	from <internal:/Users/rescueready/.rbenv/versions/3.1.2/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from bin/configure:3:in `<main>'
```

esmale:
The solution appears to be to run `bundle install` prior to running `bin/configure` if you don't have the colorize gem installed already.

esmale:
Happy to open a PR and update the documentation if that's the desired fix for this problem.

morgler:
I get the same error even after running `bundle install`. When I do a `bundle exec gem list` I even see the colorize gem.

PR #226, opened 2022-06-13
Use correct colorize method

### LINKED ISSUE #223: running bin/configure results in error: undefined method 'green'
```
rescueready@RescueReady rescue_ready % bin/configure    
Bullet Train requires Ruby 3.1.2 and `ruby -v` returns ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21].
Bullet Train requires Node.js 18.3.0 and `node -v` returns 18.3.0.
Homebrew is installed.
PostgreSQL 13 is not installed via Homebrew. Try running `brew install postgresql@13`.
Try proceeding without PostgreSQL 13?
y
Redis is installed via Homebrew.
Yarn is installed.
Next, let's push your application to GitHub.
If you would like to use another service like Gitlab to manage your repository,
you can opt out of this step and set up the repository manually.
(If you're not sure, we suggest going with GitHub)
Continue setting up with GitHub? [y/n]
n
Running `bundle install`.

  Using rake 13.0.6
  [...using lots of other gems too...]
  Using webdrivers 5.0.0
  Using wine_bouncer 1.0.4 from https://github.com/bullet-train-co/wine_bouncer.git
  Bundle complete! 49 Gemfile dependencies, 223 gems now installed.
  Use `bundle info [gemname]` to see where a bundled gem is installed.

bin/configure:156:in `<main>': undefined method `green' for main:Object (NoMethodError)

puts green "Running `yarn install`."
     ^^^^^

```

gazayas:
Fixes #223.

### Details
We merged #183 recently, but I missed the method on line 156... 🙇‍♂️️

### Move dependency to `bullet_train-base` gemspec?
In Issue #222, developers who don't already have colorize on their machine will get a `cannot load colorize` error. I mentioned putting `colorize` in a gemspec in PR #183, specifically for `bullet_train-base`. Since we run `bundle install` later in `bin/configure`, it would make sense to have it as a run-time dependency that's already downloaded before the developer runs `bin/configure`. So we can either...
1. Remove `colorize` from `Gemfile` here and move it to the `bullet_train-base` gemspec
2. Revert to the manual methods we had in the first place.

If option 1 seems reasonable, I'll make a PR in `bullet_train-base`.

Tagging @esmale for reporting.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -153,7 +153,7 @@ end
 puts "Running `bundle install`.".green
 stream "bundle install"
 
-puts green "Running `yarn install`."
+puts "Running `yarn install`.".green
 stream "yarn install"
 
 # This should be available now.


andrewculver:
Fixed this before noticing this PR!

gazayas:
@andrewculver Np 🙌

PR #237, opened 2022-06-15
Gitpod

existentialmutt:
This PR adds configuration for gitpod.  The provided configuration sets up a database, installs dependencies, runs bin/dev, and opens a browser preview of the running app

Simple_Browser_—_bullet_train_—_Gitpod_Code_and_rrosen_rafes-macbook-air____code_bullet_train
.


existentialmutt:
@andrewculver thinking it could be nice to have Gitpod run `bin/configure` for a new app.  Is there a good way to check if that script has already been run?

andrewculver:
Ah, let’s not! It assumes too much about why they’re booting the app. Want to be able to use this for actually developing BT as well.

config/environments/development.rb
```diff
@@ -85,6 +85,7 @@
 
   # allow users to access this application via the configured application domain.
   config.hosts << default_url_options_from_base_url[:host]
+  config.hosts << /.*\.gitpod\.io/
```
andrewculver, line 88: Is there any way of detecting whether we are running on a gitpod?

existentialmutt: Yes I believe there are environment variables that get set.  I'll add a check for them.

PR #244, opened 2022-06-17
Add setup script to update BASE_URL with current ngrok url

cnorm35:
With the price increase of the ngrok paid plan, I thought it would be nice to add some ways to leverage the free account easier.

Running ngrok in `Procfile.dev` may be too opinionated but wanted to include it for a working example.

new script `bin/set-ngrok-url`

- Checks for running ngrok instance
- Outputs warning message if ngrok not running
- If ngrok is currently running, grab the current public URL
- If the current value of `BASE_URL` is different from the ngrok URL, update `BASE_URL` to current ngrok URL
- Output success message with URL (handy for command + click in the console to open)
- Whitelists ngrok URLs for hosts config

A script seemed the easier to evaluate if something like this would be useful.  I've also done something similar in a rake task to have the ability to check the current environment before executing.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -3,3 +3,4 @@ worker: bundle exec sidekiq -t 25
 js: THEME="light" yarn build --watch
 light-css: yarn light:build:css --watch
 light-mailer-css: yarn light:build:mailer:css --watch
+ngrok: ngrok http 3000
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -3,4 +3,4 @@ worker: bundle exec sidekiq -t 25
 js: THEME="light" yarn build --watch
 light-css: yarn light:build:css --watch
 light-mailer-css: yarn light:build:mailer:css --watch
-ngrok: ngrok http 3000
+# ngrok: ngrok http 3000
--- a/README.md
+++ b/README.md
@@ -96,6 +96,10 @@ If you're looking contribute to Bullet Train, you should "Fork" this template re
     bin/dev
     ```
 
+    > [!NOTE]
+    > Optional: If you have [ngrok](https://ngrok.com/) installed, uncomment `ngrok: ngrok http 3000` from `Procfile.dev` and run
+    > `bin/set-ngrok-url` to set `BASE_URL` to a publically accessible domain.
+
 8. Visit http://localhost:3000.
 
 ---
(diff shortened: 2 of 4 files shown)


andrewculver:
@cnorm35 This is awesome! I agree we want to do something with tunneling in `bin/dev` and `Procfile.dev`. Leave it with me for a bit!

jagthedrummer:
@andrewculver, what are your thoughts on this? Should we resolve conflicts and get it merged?

cnorm35:
Happy to resolve conflicts and make sure this still works as intended if you all think it's useful.  I know ngrok has had some changes since this was originally opened but I believe everything still works.

jagthedrummer:
@cnorm35 Overall it seems like a great change to me. The only suggestions I would make are to _maybe_ have ngrok commented out by default in `Procfile.dev` so that we're not forcing it on people, and it would be nice to add something to the docs about the `set-ngrok-url` script (not sure where exactly). Otherwise people will have to stumble upon it on their own and figure out (or guess) when/why to use it. If you can make those changes (and resolve the conflicts) I'm comfortable making the executive decision to merge it.

cnorm35:
Sounds good @jagthedrummer, I'll get everything updated and working.  

Totally agree with having it commented out initially in the Procfile also. If that ngrok process fails, it kills all the other processes so can be a pain if you're not expecting it.

cnorm35:
@jagthedrummer Just resolved the conflicts and made a change for the updated response `ngrok` is sending and rescuing from missing file error if `config/application.yml` isn't set just in case.

Added something to the README about the script and Procfile.  Let me know if you think that can be improved or moved to a better spot.  Including a screenshot for how the updated README appears.

Screenshot 2023-08-08 at 1 16 35 PM

jagthedrummer:
@cnorm35 I think the location of that note is great. Should we also mention that you need to re-run the script every time you stop and restart the server? (Is that accurate? And/or are there other situations where you'd need to re-run the script?)

Also, I hate to have to ask, but would you mind merging in the current state of `main` to your branch? I had let something sneak through earlier today that was causing tests to fail in `main` (🤦‍♂️) and your branch is failing in the same way now.

cnorm35:
>  Should we also mention that you need to re-run the script every time you stop and restart the server? (Is that accurate? And/or are there other situations where you'd need to re-run the script?)

If `ngrok` is running via the Procfile.dev, it will have a new URL when it's re-started so that script will need to be re-ran.  I'll add something generic like 'when ngrok is restarted....' and pull in those fixes from main.

cnorm35:
@jagthedrummer pulled in that fix from `main` and everything is passing now.  Here's an updated screenshot of the additional lines I added to the README.  Let me know if there's anything else you think should be updated!

Screenshot 2023-08-08 at 1 39 05 PM

jagthedrummer (approved):

jagthedrummer:
Thank you @cnorm35, this is looking great!

PR #252, opened 2022-06-21
Update cloudinary widget

### LINKED ISSUE #219: Cloundinary UploadWidget outdated
Issue: Whilst trying to use the `User` profile image uploader (locally), the uploader pops up as expected, but the spinner spins perpetually 🙃. 
Ejecting the core gem and updating the cloudinary widget corrected all the things.

Update this: `<script src="//widget.cloudinary.com/v2.0/global/all.js" type="text/javascript"></script>`
to this: `<script src="https://upload-widget.cloudinary.com/2.1.15/global/all.js" type="text/javascript"></script>`
from here `https://github.com/bullet-train-co/bullet_train/blob/main/app/views/layouts/_head.html.erb#L20`.

gazayas:
Closes #219.

# Before
Screenshot from 2022-06-21 20-04-30

# After
Screenshot from 2022-06-21 20-05-43

# Bug
However, when actually trying to save the resource after uploading the image, I get the following error (I named the model `Pic`):
```
ActiveSupport::MessageVerifier::InvalidSignature in Account::PicsController#create
```
Maybe I'm not creating the model correctly, but if this is legitimately a bug I can look into this too and open a separate PR.

Thanks for reporting, @malachaifrazier!

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -17,7 +17,7 @@
 <meta name="turbo-cache-control" content="no-cache">
 
 <% if cloudinary_enabled? %>
-<script src="//widget.cloudinary.com/v2.0/global/all.js" type="text/javascript"></script>
+<script src="https://upload-widget.cloudinary.com/2.1.15/global/all.js" type="text/javascript"></script>
 <% end %>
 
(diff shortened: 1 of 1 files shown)


gazayas:
I was getting the error because I was using `attachment` instead of `string`. Using `string` worked just fine 👍

PR #269, opened 2022-06-29
Adjust system test for update to `invitation_uuid` logic in base package

gazayas:
Joint PR → [bullet_train-base#65](https://github.com/bullet-train-co/bullet_train-base/pull/65)

## Details
I'm assuming this check should be here when signing up, so I've simply added the button click to keep things moving instead of adjusting the code over in `bullet_train-base`.

With the new code in the joint PR, the user will now be prompted if they want to create a new account since the invitation was technically sent to a different email, or if they want to continue joining the new team with the email they just signed up with:
```
"This invitation was emailed to takashi.yamaguchi@gmail.com, but you're currently signed in as takashi@yamaguchi.com.
You can either:
Accept the invitation below and you'll join the team as takashi@yamaguchi.com.
Join The Testing Team
OR
Sign out below, re-click the link in your email, and either sign into another account or create a new account."
```

If this isn't what we should be doing by design, then I can look over the original PR and make some adjustments.


andrewculver:
@gazayas Test failures look legit.

PR #271, opened 2022-06-30
Delete unneeded comment (xray-rails) in Gemfile

gazayas:
I'm assuming we don't need these lines anymore since we merged in [nice_partials#28](https://github.com/bullet-train-co/nice_partials/pull/28)

Thanks @kaspth!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -172,9 +172,6 @@ gem "active_hash", github: "bullet-train-co/active_hash"
 # TODO Have to specify this dependency here until our changes are in the original package or properly forked.
 gem "wine_bouncer", github: "bullet-train-co/wine_bouncer"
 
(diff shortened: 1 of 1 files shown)


kaspth:
Don't think we can do this yet, it looks like the compatibility fix PR is still open on xray-rails https://github.com/brentd/xray-rails/pull/108

gazayas:
Okay, I guess I'm just confused as to why it's commented out. Are we leaving it just in case developers want to use xray?

andrewculver:
@gazayas Ah, yes, this is safe to remove because we got rid of X-ray as a dependency! We just use the regular Rails annotations now!

gazayas:
@andrewculver Thanks for clarifying 🙌

PR #303, opened 2022-07-17
Provide consistent guidelines for each question in the bin/configure script

etagwerker:
Hi there,

Small language improvement. I figured it would be best to make it consistent across all questions in the configure script.

Please check it out.

Thanks!

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -57,7 +57,7 @@ if actual_ruby.include?(required_ruby)
   puts message.green
 else
   puts message.red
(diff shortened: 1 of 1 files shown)


andrewculver:
@etagwerker Hey! Thanks for this! The intention of the `[y]` is to indicate what the default will be if you just hit <kbd>Return</kbd>. For some reason I thought this was a bit of a convention in UNIX shell scripts, although in this example it's done with parenthesis:

Screen Shot 2022-07-26 at 2 49 48 PM

Either way, I'm open to any suggestions on how we can make this more user-friendly.

PR #338, opened 2022-08-02
Only setup `PartialTest` on one node in parallel tests

gazayas:
Addresses the TODO in `test/bin/setup-super-scaffolding-system-test`.

If we want to keep parallelism to 7, I can shift things around and see how we can fit it into the space we have.


gazayas:
As you can see with the last build, we are now skipping `PartialTest` on all other nodes
Skipping PartialTest

gazayas:
Closing in favor of #343.

PR #343, opened 2022-08-05
Cleaning up Super Scaffolding Partial Tests

gazayas:
At first I set out to simply clean up the test files that were being generated in `tmp/storage` whenever running the Super Scaffolding `TestFile` test, but it turned out there was some more cleanup to get taken care of, so here are the different things I addressed.

## Changes
1. Remove the `foo.txt` file that’s generated whenever we run `super_scaffolding_system_test.rb`
2. Move all the necessary logic to `PartialTest` and delete `TestFile` (I wrote `TestFile` before `PartialTest` was merged in, so these were separated before)
3. Make sure `PartialTest` runs in parallel

I made sure PartialTest ran in parallel in #338, but I figured there would be conflicts so I just decided to take care of it here.


andrewculver:
@gazayas Can you resolve the conflict here?

gazayas:
Closing in favor of #654.

PR #365, opened 2022-08-17
Re-enable overmind

gazayas:
I'm curious if this is an M1 issue or if it has actually been fixed, because I updated tmux to 3.3a on my mac (not M1) with brew and it's working for me. Thought I'd open this PR in case it started working for anyone else too.

tmux 3 3a

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -4,7 +4,6 @@ bin/link
 
 if [ -x "$(command -v overmind)" ]
 then
-  echo "RUNNING OVERMIND"
   overmind start -f Procfile.dev "$@"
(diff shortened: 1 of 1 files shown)


gazayas:
There's also a newer comment over in the issue in the comment I deleted which suggests running this:

```
> brew uninstall --ignore-dependencies tmux
> brew install tmux
```

But I feel like it would be more of a hassle to require our developers to do this if it's not working, so if it's not working on M1s, it's probably better to just wait as opposed to trying to run it in something like `bin/configure`.

PR #368, opened 2022-08-18
Document specific details about Doorkeeper grant flows

gazayas:
## TL;DR
If there's no specific reason to use the `password` grant flow, we probably shouldn't use it because of the security risks that come along with it.

## Details
If no grant flows are specified, Doorkeeper automatically sets these values to `authorization_code` and `client_credentials`, so these should be the default values.

`implicit` and `password` have pitfalls to look out for, so I read up on the information provided by the links and paraphrased it to make it somewhat shorter.

The password grant seems to be the most high-risk, so if we don't have a need for it we can probably remove it from `grant_flows`, but I'll leave that up to your discretion.

### DIFF
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -358,7 +358,20 @@ Doorkeeper.configure do
   #   http://tools.ietf.org/html/rfc6819#section-4.4.2
   #   http://tools.ietf.org/html/rfc6819#section-4.4.3
   #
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
This is a good change. I opened this issue to track/discuss the idea that we should remove `password` and maybe `implicit` from the default `grant_flows` list. https://github.com/bullet-train-co/bullet_train/issues/820

PR #431, opened 2022-09-19
add asdf option to bin/config and update postgres to 14

### LINKED ISSUE #409: Using `asdf` for managing ruby and node versions
Hello i am new to bullet train and was wondering if there has been any discussion on using asdf to manage ruby and nodejs versions?

I noticed the `bin/configure:93` checks to see if node is installed with brew using the `check_package` which function runs the `not_installed?` function.

Any input would be great!

[asdf-vm.com](https://asdf-vm.com/)
[github.com/asdf-vm/asdf](https://github.com/asdf-vm/asdf)

iseth:
closes #409


jagthedrummer (approved):
Personally I'm a fan of `asdf` and think this would be a good change. @andrewculver, @gazayas what do y'all think?

Also worth mentioning that the PR title says that it updates postgres to 14, but it looks to me like that's already been done outside of this PR. So I think this PR is _only_ about `asdf` now.

jagthedrummer:
@iseth There seem to be some merge conflicts on this PR now. Can you take a look and resolve them?

gazayas:
> Personally I'm a fan of `asdf` and think this would be a good change. @andrewculver, @gazayas what do y'all think?
> 
> Also worth mentioning that the PR title says that it updates postgres to 14, but it looks to me like that's already been done outside of this PR. So I think this PR is _only_ about `asdf` now.

I'm down to use `asdf`, I think it also came up in conversation a while back with another member on the team.

bin/configure
```diff
@@ -64,7 +64,10 @@ end
 
 # Unless the shell's current version of Ruby is the same as what the application requires, we should flag it.
 # rbenv produces strings like "3.1.2" while rvm produces ones like "ruby-3.1.2", so we account for that here.
-required_ruby = `cat ./.ruby-version`.strip.gsub(/^ruby-/, "")
+asdf_version = `cat ./.tool-versions`.lines.find { |i| i.include?("ruby") }.gsub(/\b\w*[ruby]\w*\b/i, "").strip
+rvm_version = `cat ./.ruby-version`.strip.gsub(/^ruby-/, "")
```
gazayas, line 68: Can we potentially leave the variable name here as `required_ruby`, or maybe use a switch statement for the different types of version managers? I personally use rbenv, and `.ruby-version` only has the version number (i.e. - `3.2.2`) which shouldn't have any bearing on `gsub(/^ruby-/, "")`. What do you think?

jagthedrummer:
@iseth Thanks a bunch for the PR, but in the time between when you opened it and now much of what you changed has already been incorporated into `bin/configure`. After I rebased this branch and removed duplication (or duplication of effort) the only thing left was the introduction of the `.tool-versions` file. Since `asdf` will look at both `.ruby-version` and `.nvmrc` that seems like unnecessary duplication. I'm going to close this, but if anything is still giving you trouble please do open another PR. (And we'll try to get to it sooner than this last time. 😬)

PR #435, opened 2022-09-21
Test selecting time-zone during signup

pascallaliberte:
The process of selecting a time-zone during sign up is cumbersome because [too many clicks are needed to use type-ahead search for `super_select` field that allow a single selection](https://github.com/bullet-train-co/bullet_train-fields/issues/35).

Also, the existing `select2_select` helper only works with `super_select` fields that allow multiple selections.

This PR:

* Makes the `select2_select` helper work with `super_select` fields that allow only a single selection. It simulates typing into the un-opened field and continues to type with the super_select dropdown opened and the input field in focus.
* Adds a step to `authentication_test` to select the Time Zone during sign up.

Co-dependent PRs (required for the test to pass):
* https://github.com/bullet-train-co/bullet_train-themes/pull/21
* https://github.com/bullet-train-co/bullet_train-fields/pull/36

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -187,10 +187,12 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     field = find("label", text: /\A#{label}\z/)
     field.click
     "#{string}\n".chars.each do |digit|
-      within(field.find(:xpath, "..")) do
-        find(".select2-search__field").send_keys(digit)
-      end
+      active_element.send_keys(digit)
     end
+  end
+  
+  def active_element
+    page.driver.browser.switch_to.active_element
   end
 
   # https://stackoverflow.com/a/50794401/2414273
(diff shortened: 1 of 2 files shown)


pascallaliberte:
With the associated PRs, we get this smooth, smooth time-zone-picking flow:

CleanShot 2022-09-21 at 14 26 20

jagthedrummer:
@pascallaliberte This looks good to me, and it looks like the dependent PRs were replaced by one in `core` that has already been merged. So is it accurate that this one should be ready to merge now? If so we should re-trigger CircleCI, which I don't have permission to do, so we can make sure that tests are passing. (It looks like they failed previously, but it's been so long that Circle doesn't seem to have those test runs available to inspect anymore.)

jagthedrummer:
@pascallaliberte I saw that there was a linter error in Circle, so I fixed that, which got a new Circle run to kick off. But now there seems to be a legitimate failure related to these new tests. Can you take a look?

jagthedrummer:
@pascallaliberte Is this still something you'd like to get merged?

pascallaliberte:
@jagthedrummer yes, I'll scrub and polish it next week

PR #436, opened 2022-09-21
Add window.Stimulus for help debugging, even in production

pascallaliberte:
To help debug Stimulus controllers, this PR adds `window.Stimulus` and a comment to hint on two common ways to use it (debugging events and finding the list of loaded controllers).

It's arguably safe to use in production, unless Stimulus controllers expose methods that access resources that aren't protected by the application, which should be dealt with at the application level.

### DIFF
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -9,10 +9,11 @@ import consumer from '../channels/consumer'
 
 const application = Application.start()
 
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
This seems like a good change to me, but I'm not _super_ familiar with Stimulus. @andrewculver do you see any reason not to merge this?

kaspth (approved):

PR #457, opened 2022-10-04
Add `current_version_number` as attribute for API

gazayas:
Details in main PR:
- https://github.com/bullet-train-co/bullet_train-base/pull/127

Joint PR
- https://github.com/bullet-train-co/bullet_train-api/pull/34

### DIFF
--- a/config/initializers/api.rb
+++ b/config/initializers/api.rb
@@ -1,1 +1,2 @@
 BulletTrain::Api.current_version = "v1"
+BulletTrain::Api.current_version_number = 1 # Added for performance reasons.


jagthedrummer:
@gazayas Is this PR still something we need? It looks like the dependent PRs were for repos that have since been moved into `core`. And I doesn't look like those changes are present in `core` and I don't find a new PR in `core` to make those changes.

gazayas:
@jagthedrummer Yes, we use the following now, so these PRs are obsolete:

```ruby
controller = "Api::V#{api_version}::ApplicationController".constantize.new
```

However we might be able to glean performance changes from these PRs, so if there are any other performance issues we should address I can look into them. I say that because the [original TODO](https://github.com/bullet-train-co/bullet_train-base/blob/4e6ae5c3a346e0c7f2c17823911ab561ea0d913c/app/models/concerns/records/base.rb#L86) stated this:

```
# TODO So many performance improvements available here.
```

I'll go ahead and close these out, but I'll make a note in case we should be addressing these performance issues.

Issue #489, opened 2022-10-25
Deploy to Heroku button not working


jot:
I get the following error message when using the Deploy to Heroku button directly from either [README.md](https://github.com/bullet-train-co/bullet_train#readme) or https://bullettrain.co/docs/heroku

`Couldn't find either the add-on service or the add-on plan of "honeybadger:free".`

### Screenshot:
<img width="653" alt="CleanShot 2022-10-25 at 13 37 55@2x" src="https://user-images.githubusercontent.com/4997/197774917-f87fb7b6-d2b8-4ea0-8d4a-34421adb5ed2.png">

andrewculver:
@jot Confirmed with another user yesterday that this is working now! Thanks for reporting this!

PR #490, opened 2022-10-25
Update app.json to reference correct honeybadger addon name

jot:
Issue described in #489.

It turns out the free version of the addon is now(?) called `basic` rather than `free`


andrewculver:
@jot Thanks for this!

PR #497, opened 2022-10-29
Add system tests for Action Models

### LINKED ISSUE #492: Add system tests for Action Models.
You can see an example of how we set up and then perform a system test for Super Scaffolding in the following three files:

- CI Config: https://github.com/bullet-train-co/bullet_train/blob/main/.circleci/config.yml#L134
- Setup: https://github.com/bullet-train-co/bullet_train/blob/main/test/bin/setup-super-scaffolding-system-test
- Test: https://github.com/bullet-train-co/bullet_train/blob/main/test/system/super_scaffolding_test.rb

We need to do the same thing for each of the Action Model scaffolders, confirming that when `bullet_train-action_models` is installed, you can generate each of the different types of actions and they function in the application as expected.

We should run this on CI as a separate task from the Super Scaffolding tests. However, we should probably group them together in a single CI task similar to how all the Super Scaffolding tests are together in a single task.

The reason we add these tests in this repository is so folks building with Bullet Train can confirm that nothing they've done in their local applications has broken the ability to generate and execute actions.

gazayas:
Closes #492.

I'm starting with `targets-many` so we can take things one step at a time.

## Writing custom actions for the `perform_on_target` method
I mulled over this one a bit, and I chose to write the following line in the setup script via `sed` for speed:
```
project.update(name: "#{project.name} archived")
```

If the purpose of these tests is to check if the stock Action Model functionality is working properly, then this setup shouldn't be a problem. Do we want developers to edit this to test their own actions though? I'm assuming that's not the case, but if it is, we could use `File` to show how we're defining the action explicitly within the system test itself, although the tests would be slower. I'll stick with the one system test file I have here though since I'm guessing we just want to test the stock functionality, similar to the Super Scaffolding System test.

## Branch to add for tests to pass
I have a branch on the Action Models repository that will let these tests pass
- https://github.com/bullet-train-pro/bullet_train-action_models/pull/15

I'll see what I can do with CircleCI.


gazayas:
Just added a test for `targets-one`, but running into an `inverse_of` error so I left a TODO there for now. Planning on working on `targets-one-parent` next.

gazayas:
@andrewculver `targets-many` wasn't too much of a problem to get going, but without going into too much detail here, I've skipped the bulk of the tests. I've decided to cut the PR here because as far as this PR, I don't think there's much we can do unless we address a couple of the issues I opened up over in Action Models.

Here are the actions I've tested for:
1. `targets-many`
2. `targets-one`
3. `peforms-import`

Sifiting through `targets-one-parent` and the brief conversation we had about testing it, this setup currently makes the most sense to me.

## Conditional Workflows
I tried to get CircleCI set up so that it only runs these tests when we set a custom parameter to `true`. The [when](https://circleci.com/docs/configuration-reference/#the-when-attribute) conditional seemed really helpful, but I was just having more problems than necessary trying to set it up, so I dropped the idea and, although it's not ideal, resorted to having the developer uncomment the test in case they want to run it on CircleCI.

## More work to do
If need be, I can look at these tests again soon if some things get changed around in Action Models.

PR #506, opened 2022-11-09
Add incoming webhooks to API routes

gazayas:
Main PR
- https://github.com/bullet-train-co/bullet_train-incoming_webhooks/pull/2

## Details
I added these namespaces explicitly so we can add newly super scaffolded incoming webhooks with ease.

### DIFF
--- a/config/routes/api/v1.rb
+++ b/config/routes/api/v1.rb
@@ -4,6 +4,11 @@ extending = {only: []}
 
 shallow do
   namespace :v1 do
+    namespace :webhooks do
+      namespace :incoming do
+      end
+    end
+
     # user specific resources.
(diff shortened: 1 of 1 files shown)


andrewculver:
@gazayas Incoming webhooks shouldn't be under `/api/v1/webhooks/incoming`. They should just be `/webhooks/incoming`.

andrewculver:
Oh, I get it now! Yes, I can see why you thought this was needed, but it's actually not needed, thankfully! We don't need people to be able to see incoming webhooks via API.

gazayas:
No problem! I wasn't entirely sure so I added it just in case. Glad the new scaffolder made it over the finish line!

PR #519, opened 2022-11-28
507 Re-enable 2FA

### LINKED ISSUE #507: 2FA should be re-enabled.
We had this enabled previously, but we disabled it when upgrading to Rails 7. You can see where we were including it at https://github.com/bullet-train-co/bullet_train-base/blob/main/bullet_train.gemspec#L46 . Would love to see this working again.

bhumi1102:
Code Changes:
- Add `otp_secret` column to `users`. This is the new column require for devise_two_factor 5.0 
- Add Rails encryption credentials. Required by devise_two_factor
- Add back the two 2FA gems

Testing Done:
- Setup 2FA locally (create rails encryption credentials, env var, etc.)
- Create a new user, login, click 'Enable' in Account Details
- Scan QR Code with Google Authenticator App on phone
- Logout and attempt to Sign In with the same user
- Before this change, Sign In unsuccessful with 401 from POST to `/users/sign_in`.
- After this change, Sign In with 2FA code from Authenticator app results in a successful sign in!

Note: follow-on change to add a `legacy_otp_secret` method will be in the bullet_train-base gem. 

closes #507


eadz:
@bhumi1102 thanks for looking into this! 

I'm not sure why `config/credentials.yml.enc` is committed here? The only way to decrypt it is with the master.key. 

I believe this file results from setting up rails encrypted db using `bin/rails db:encryption:init` in which case it should be created by the developer. 

I think we should potentially add an environment variable 'TWO_FACTOR_ENABLED', and use that instead of 'TWO_FACTOR_ENCRYPTION_KEY' to decide if two factor is enabled, since TWO_FACTOR_ENCRYPTION_KEY is ignored(?) in 5.0?

Also do we want to remove line 8 here: https://github.com/bullet-train-co/bullet_train-base/blob/a5994afb09b6f5a2f6fab649c09d49c80ae8dd90/app/models/concerns/users/base.rb#L8 ?

bhumi1102:
@eadz hi there 👋  nice to meet you! Thanks for taking a look..
- yup removed the encrypted credentials.yml
- added use of TWO_FACTOR_ENABLED instead of the indirect use of the legacy encryption one. I like it, it's clearer. (that will show up in https://github.com/bullet-train-co/bullet_train-base/pull/142 though.
- exactly, the otp_secret_encryption_key is needed for legacy encryption. Agree about adding a note - I wrote some dev docs that includes this (sent to @andrewculver to find a home for it).

andrewculver:
@bhumi1102 So sorry for the trouble, but I was wrong. We do not need to maintain backwards compatibility. We can lean into Rails encrypted secrets. Also, I'd suggest we don't need `TWO_FACTOR_ENABLED` at all. Instead, let's just assume that if the key is set via encryption credentials, the feature is enabled. If we want to add something for disabling it later when the key is in play, we can do that later, but we won't do it via ENV vars.

eadz:
> let's just assume that if the key is set via encryption credentials,

The key is not specific to 2fa, it's for all rails encrypted attributes. Given how 2fa is everywhere now, I think we could enable it by default as a good security practice, (users still have to opt-in) and have a way to opt out?

andrewculver:
Oh, I missed that! Yeah, I think we should enable 2FA by default then.

andrewculver:
OK, sorry, dug into this a little more. I didn't realize we weren't creating a separate key for 2FA. That's fine, we can depend on ActiveRecord encrypted attributes. However, this is an additional getting started step for folks, and I don't want things to error out just because they haven't finished configuring encrypted attributes, so we should still check whether ActiveRecord encrypted attributes are configured before trying to enable 2FA by default. But if and when ActiveRecord encrypted attributes are configured, then 2FA should be enabled by default.

bhumi1102:
@andrewculver thanks and it's no trouble! I like the decision of forging forward. That makes things cleaner.

So let me address 1. the idea of 'enable 2FA by default' and 2. concern about not wanting things to error out during initial BT setup for users.

For 1. It still makes sense to have end-users of BT apps opt-in to 2FA by clicking an 'Enable' button in settings. And here's why - when users click 'Enable' that is when the devise-two-factor gem does its thing, creating the otp secret as well as otp back codes and presenting a QR to the user in the UI. So this step involves user interaction and makes sense to keep intentional. In terms of 'enable 2FA by default' in for BT users (not end-users) yes we *are* doing that by including the gem, adding the correct DB column migration. So after this is merged, support for 2FA is back for BT. (minus one thing, which I'll talk about next).

2. Desire to BT initial setup and keeping that smooth and not have things error out due to 2FA. I share this. With this PR nothing should error out. *When* they click 'Enable' and they don't have rails encryption setup, that's when there will be error asking to configure rails encryption. *But* we can get ahead of this and ask BT users to configure rails encryption during initial 'getting started' setup. It's a simple one line thing (I'm happy to write/update docs clearly stating this).

Let me know if that make sense or if you have any more questions Andrew.

bhumi1102:
Oh and about the encryption keys. Here's the before and after story with devise-two-factor(D2F). D2F previously used attr_encrypted gem to transparently encrypt and decrypt the 3 `encrypted_otp_*` attributes  on `user` model. The latest version of D2F deprecates using attr_encrypted in favor for rails active record encryption, now that it's available in rails.

> I didn't realize we weren't creating a separate key for 2FA. That's fine, we can depend on ActiveRecord encrypted attributes.

We using rails AR encryption has prescribed in the D2F gem's readme (image below). We *are* also creating a secret for 2FA and storing it in the column `otp_secret` on `users`. All from the D2F gem's docs. (Previous version of D2F stored these 2FA secrets in 3 columns, with a salt and what not). *Then* we are using rails AR encryption to transparently encrypt/decrypt those attributes on the model. 

Hope that helps! 
Screen Shot 2022-12-07 at 2 30 04 PM

andrewculver:
@bhumi1102 Sorry, I should have been more clear. 🙇🏻‍♂️ I was not referring to having 2FA enabled by default for _users_, but _developers_. Like, the application should have a 2FA feature by default. (This is a common type of misunderstanding when talking about Bullet Train stuff, I should have been more careful.)

bhumi1102:
Haha, all good. I had feeling so yeah for *developers* we are enabling 2FA support with this PR basically. As I said "In terms of 'enable 2FA by default' in for BT users (not end-users) yes we are doing that by including the gem, adding the correct DB column migration." plus adding a step to configure rails AR encryption keys (as required for D2F)

bhumi1102:
Resolved merge conflicts and CI is green now also 😄 

one last thing I was thinking about, the `TWO_FACTOR_ENABLED` env var. We *could* do away with it I guess, though the method `two_factor_authentication_enabled?` is used in several places in the sessions new view to decide whether to display the UI to collect the otp attempt code from the user, etc. Now that DF2 gem uses rails encryption keys, we would have to *infer* that 2FA is enabled based on rails encryption keys being present. But since rails encryption is general purpose thing, we just happen to use it for 2FA, that wouldn't be accurate. And would be surprising (and bad) to the BT developer and end-user if adding rails encryption for unrelated thing suddenly starts asking users for otp codes during login. 

So I would lean towards keeping the env var `TWO_FACTOR_ENABLED` as it's clearer (and l like the earlier suggestion to rename it from `TWO_FACTOR_ENCRYPTION_KEY` as that is no longer used and misleading). The only downside is that it needs to be set by the BT developer to enable 2FA (and we can add that to docs)?

If there is anything else that we need to sort out before merging, do let me know!

andrewculver (approved):

PR #521, opened 2022-12-01
[RFC / WIP]: Rework bundling

KonnorRogers:
Some notable things:

- Add code splitting
- Allow multiple entrypoints which can share code.
- Remove `build:light` for JS and move it to an environment variable (happy to change the ENV var)

### DIFF
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -49,7 +49,7 @@ const ImportGlobPlugin = () => ({
 
 let themeFile = ""
 if (process.env.THEME) {
-	themeFile = execSync(`bundle exec bin/theme javascript ${process.env.THEME}`).toString().trim()
+  themeFile = execSync(`bundle exec bin/theme javascript ${process.env.THEME}`).toString().trim()
 }
 
 // Could also swap to packs?
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -73,8 +73,10 @@ require("esbuild").build({
     global: "window"
   },
   bundle: true,
-  splitting: true,
-  format: "esm",
+  // ESM + Splitting will only work if the script is type="module"
+  // splitting: true,
+  // format: "esm",
+  format: "iife",
   sourcemap: true,
   outdir: path.join(process.cwd(), "app/assets/builds"),
   loader: {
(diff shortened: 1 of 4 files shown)


esbuild.config.js
```diff
@@ -47,18 +47,36 @@ const ImportGlobPlugin = () => ({
...
+    "application": path.join(process.cwd(), "app/javascript/application.js"),
+    "intl-tel-input-utils": path.join(process.cwd(), "app/javascript/intl-tel-input-utils.js"),
+    [`application.${process.env.THEME}`]: themeFile,
+},
   define: {
     global: "window"
   },
   bundle: true,
+  splitting: true,
+  format: "esm",
```
KonnorRogers, line 77: I don't know where the script tags for these built JS files go, but ideally it should look like this: `<script src="application.js" type="module">`

andrewculver:
@gazayas Can you double check that theme ejection still works after this?

package.json
```diff
@@ -32,9 +32,8 @@
     "tailwindcss": "^3.2.3"
   },
   "scripts": {
-    "build": "node esbuild.config.js",
-    "build:css": "bin/link; yarn light:build; yarn light:build:css; yarn light:build:mailer:css",
-    "light:build": "esbuild `bundle exec bin/theme javascript light` --bundle --sourcemap --outdir=app/assets/builds --loader:.png=file --loader:.jpg=file",
+    "build": "THEME=\"light\" node esbuild.config.js",
+    "build:css": "bin/link; THEME=\"light\" yarn build; yarn light:build:css; yarn light:build:mailer:css",
```
jagthedrummer, line 36: @KonnorRogers, I'm sorry to comment on such an old PR, but I'm wondering if you might happen to remember why you added `yarn build` as a part of `yarn build:css` here. It results in `yarn build` being run twice when we call `rails assets:precompile`. It gets called once directly by the `precompile` task, and then gets called a second time as part of `build:css`.

KonnorRogers: No worries. Umm...oversight maybe? I can't remember if there was some issue with needing to build in the CSS because of race conditions in the Procfile...it's been a while 😅

KonnorRogers: Either that, or I just blindly followed what we there before and didn't realize it...really not sure tbh...

PR #529, opened 2022-12-07
Upgrade to Sidekiq 7, remove initializer

gazayas:
I noticed that the depfu PR to upgrade to Sidekiq 7 was failing, so I looked into it and found out that the Redis connection pool value is now set automatically with Sidekiq 7.

Since we were setting other values like `url` and `ssl_params` in the initializer I thought we'd still need to set them somehow, but it seems like Sidekiq 7 takes care of all of the defaults for us. There are some more details in [this Sidekiq issue](https://github.com/mperham/sidekiq/issues/5608). The author of the issue was setting `redis_conn` similar to what we had in our initializer, but ultimately this was the response from the maintainer:
> You can completely remove your initializer and it should just work.

I didn't want to just get rid of everything though, especially if production requires a proper `REDIS_URL`, but after debugging a bit it looks like Sidekiq really does take care of all of the necessary settings out of the box.

For example, if we set `REDIS_URL` in `config/application.yml` it will use that, and if not it will default to `redis://localhost:6379`. Here are the results from prying into a unit test (you can see at the end that the client has been changed):
```
# After setting REDIS_URL to `redis://test:6379`...

[1] pry(#<AuthenticationSystemTest>)> Sidekiq.redis {|redis_config| p redis_config}
2022-12-07T07:29:00.040Z pid=16350 tid=7ey INFO: Sidekiq 7.0.2 connecting to Redis with options {:size=>5, :pool_name=>"internal", :url=>"redis://test:6379"}
#<Sidekiq::RedisClientAdapter::CompatClient:0x00007f5b8bff9e00 @client=#<RedisClient redis://test:6379/0>, @_pipeline_class=Sidekiq::RedisClientAdapter::CompatClient::Pipeline>
=> #<Sidekiq::RedisClientAdapter::CompatClient:0x00007f5b8bff9e00
 @_pipeline_class=Sidekiq::RedisClientAdapter::CompatClient::Pipeline,
 @client=#<RedisClient redis://test:6379/0>>
```

## `configure_server` not working
I'm not sure if we'll need the initializer in the future, but I noticed that `configure_server` wasn't running. As a side note, `configure_client` still works.
```ruby
Sidekiq.configure_server do |config|
  puts "This doesn't run..."
end
```


andrewculver:
@gazayas Do you mind resolving the merge conflict here? Sorry for the delay merging this.

PR #533, opened 2022-12-12
Add applications controller test

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train-api/pull/45

The test for making sure applications are created when accessing `/testing/provision`: https://github.com/bullet-train-co/bullet_train-api/issues/42


jagthedrummer:
@gazayas Do we still need this? If so, can yo help me understand why someone might want to provision a platform application this way during tests rather than just by creating it directly during the setup phase?

gazayas:
I mainly wrote this in response to https://github.com/bullet-train-co/bullet_train-api/issues/42 so we could have tests of some kind, but I agree that we could verify the logic in the setup phase so I'm currently impartial to how we handle the tests here. Should I use the endpoint in the [application platform test](https://github.com/bullet-train-co/bullet_train/blob/main/test/system/application_platform_test.rb), or should we just close this one out?

jagthedrummer:
@gazayas I think there may be just a little more to do here. This branch is failing CI because of the webdrivers problem that we fixed here: https://github.com/bullet-train-co/bullet_train/pull/819.

To see if I could get it ready to merge I pulled your branch, rebased it against main, and then opened a new PR here: https://github.com/bullet-train-co/bullet_train/pull/821

_That_ branch is failing CI with an error about `No route matches [GET] "/testing/provision"`. I get the same error locally when running the test if `TESTING_PROVISION_KEY` is not set.

So the `else` branch doesn't seem to be a valid test. It looks that that route simply doesn't exist if the key isn't set.

For CI purposes I think we'd want to test both branches, but I'm not sure it's worth setting up an entire new test phase just for that (like we did with `HIDE_THINGS`).

Do you think it would reasonable to explicitly set/unset `ENV['TESTING_PROVISION_KEY']` inside the test suite? I think it might be the least worst option, as long as we preserve the original value and reset it after that test is over. What do you think?

gazayas:
@jagthedrummer Since I duplicated the environment variable check in both the routes file and the controller action, I decided to simply declare the route as is and leave the variable check in the action in https://github.com/bullet-train-co/bullet_train-core/pull/373. I also edited the tests here a bit.

That should take care of it, but if there's anything I'm missing I'd be glad to look over it again!

jagthedrummer (approved):

jagthedrummer:
@gazayas I think we should hold off on merging this one until we can get a new version of the `core` gems released. If we merged it right now we'd be setting the starter repo up to have a failing test right out of the gate for anyone who cloned it.

gazayas:
@jagthedrummer Got it, I saw that #847 was merged so I opened a new PR at #852.

PR #590, opened 2023-01-02
DRY translations.

andrewculver:


jagthedrummer:
@andrewculver, there's not much context here. Is this something we still need?

jagthedrummer:
I don't find a `TranslationsBackfiller` class anywhere in any of the BT repos, so I'm assuming that this was the first step in a work-in-progress and is not ready to merge.

jagthedrummer:
In the interest of minimizing in-flight WIP I'm going to close this. It's a very smol change that will be easy to recover if this comes up again.

PR #598, opened 2023-01-06
Tests for redirecting to show page for new records

gazayas:
These are the tests for https://github.com/bullet-train-co/bullet_train-core/pull/49.
I also fixed a small bug in the Super Scaffolding Partial system test along the way.

These fixes mostly consist of:
1. Clicking "Back" so we can continue the system tests on the model's index page.
2. Deleting unneeded clicks which had no assertions after them inside the Super Scaffolding system test.

## Changelog
I think developers will need to change their tests for Super Scaffolded models, specifically for anything generated by `test/controllers/account/scaffolding/completely_concrete/tangible_things_controller_test.rb`.


gazayas:
We changed `bin/checkout-and-link-starter-repo` recently so the tests won't pass on CircleCI, but everything is passing for me locally 👍

PR #599, opened 2023-01-06
Update README to add running setup script

siaw23:
Closes #573

PR #605, opened 2023-01-08
Update bin/configure

siaw23:
Since #599, users are expected to have already ran bin/setup. No need to repeat that

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -209,5 +209,5 @@ end
 puts ""
 puts "OK, we're done, but at some point you should edit `config/locales/en/application.en.yml`!".yellow
 puts ""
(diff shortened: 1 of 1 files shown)


vfonic:
Oh well... we've come full circle. :)

There are two of us now claiming that that last message should mention `bin/setup`. I believe the mention of `bin/setup` shouldn't have been removed.

So instead of:

```rb
puts "Next you can run `bin/setup` and then `bin/dev`.".green
# or
puts "Next you can run `bin/dev` to spawn a local instance, then navigate to http://localhost:3000.".green
```

It would probably be better to have something along these lines:

```rb
puts "Next you can run `bin/setup` if you haven't already and then run `bin/dev` to start local server at http://localhost:3000.".green
```

What do you think @siaw23?

PR #608, opened 2023-01-09
Remove uncessary nested unless block

vsnguyen:
This PR removes the nested `unless` block from `line 11`.  The `unless` from `line 7` should be enough.


andrewculver:
@vsnguyen Thanks for this!

Issue #607, opened 2023-01-09
`bin/setup` step: `config/application.rb:56: syntax error, unexpected '\n'`


Natblow:
OS: Ubuntu 20.04

When I run `bin/setup` it goes well until when it comes to prepare the database step, I have a syntax error : 

```
config/application.rb:56: syntax error, unexpected '\n', expecting '.' or &. or :: or '['
/home/natblow/ffwpu_ca/community_platform/Rakefile:4:in `require_relative'
/home/natblow/ffwpu_ca/community_platform/Rakefile:4:in `<main>'
bin/rails:4:in `<main>'
(See full trace by running task with --trace)
```

I'm not sure where it may come from. The code looks ok.
Maybe due to the upgrade ro Ruby 3.2.0 ? Edit: No, I downgraded to Ruby 3.1.2 but still same problem.

Natblow:
Turns out the module didn't have a name on line 12 of `config/application.rb`

I'm wondering why it didn't write the name from the `bin/configure` command.

PR #610, opened 2023-01-10
update `bin/configure` to prohibit empty application name

Natblow:
On my first time using of Bullet Train, I've opened an issue #607 where I could resolve later the same day. 
I must have put an empty string for some reason when prompted about the name of an application which caused that issue. 

This PR is to prevent this problem if it occurs.

Also I have added `[y/n]` on the first question prompt.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -143,6 +143,10 @@ if skip_github.downcase[0] == "y"
     `#{command} https://github.com/new`
 
     ssh_path = ask "OK, what was the SSH path? (It should look like `git@github.com:your-account/your-new-repo.git`.)"
+    while ssh_path == ""
+      puts "You must provide a path for your new repository.".red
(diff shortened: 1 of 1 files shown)


kaspth:
Sweet! I glanced over the file and it looks like `ssh_path` can have the same issue: https://github.com/bullet-train-co/bullet_train/blob/d0ea533c4bc3bc25739f69610f1b5f656587587d/bin/configure#L145

I think we should fix that one too.

It looks like several of the `ask` lines could use the `[y/n]` treatment, can we split that into a separate PR and update all of those?

Natblow:
Thanks @kaspth , those are great points! 
I now added the same restriction for the `ssh_path` value.

I've also added `[y/n]` options on all ask lines where necessary, on a separate PR #613

kaspth:
@Natblow sweeeet, thank you so much!

PR #613, opened 2023-01-10
update `bin/configure` to standardize ask prompts with `[y/n]`

Natblow:
I think it would be better to have the options `[y/n]` being asked more consistently when needed.


kaspth:
Nice!

Issue #580, opened 2022-12-28
Make drop-down menu wider?


andrewculver:
Just saw this on https://bullettrain.co:

<img width="264" alt="Screen Shot 2022-12-27 at 11 28 38 PM" src="https://user-images.githubusercontent.com/217218/209775267-fcc80d69-5048-4bde-84de-1354de7d7330.png">

jagthedrummer:
Fixed in https://github.com/bullet-train-co/bullet_train/pull/615

PR #615, opened 2023-01-11
Fix Drop Down Menu for Integration

ps-ruby:
https://github.com/bullet-train-co/bullet_train/issues/580

Issue:
image

The UI for the drop-down menu was broken as shown above because `@menu_orientation` was coming as `top` instead of `side`. The following hack has fixed the issue but I am not sure if this is the correct approach to fix this issue. 

Fixed:
image

### DIFF
--- a/app/views/account/shared/_menu.html.erb
+++ b/app/views/account/shared/_menu.html.erb
@@ -8,8 +8,11 @@
   <% end %>
 
   <% integrations = capture do %>
+    <% @last_menu_orientation = @menu_orientation %>
+    <% @menu_orientation = :side %>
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
After looking into it this seems to be exactly what we do for other menus. For example here: https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-themes-light/app/views/themes/light/menu/_account.html.erb#L19-L22

It feels a little... hacky(?) to me. But it seems to be the established pattern, so I think it's good enough to merge. If we want to clean up the pattern we can do that in a different issue/PR.

PR #630, opened 2023-01-21
Minor edit to spacing in API controller test

gazayas:
When Super Scaffolding a new model, there's a gap of extra space left before all the test's contents because the following two lines get skipped:
`test/controllers/api/v1/scaffolding/completely_concrete/tangible_things_controller_test.rb`
```ruby
unless scaffolding_things_disabled? # 🚅 skip when scaffolding.
 ...
end # 🚅 skip when scaffolding.
```

All I did was move the test contents one tab over to the left.

## Before
```ruby
class Api::V1::FoosControllerTest < Api::Test
    def setup
      # See `test/controllers/api/test.rb` for common set up for API tests.
      super

      @foo = build(:foo, team: @team)
      @other_foos = create_list(:foo, 3)
```

## After
```ruby
class Api::V1::FoosControllerTest < Api::Test
  def setup
    # See `test/controllers/api/test.rb` for common set up for API tests.
    super

    @foo = build(:foo, team: @team)
    @other_foos = create_list(:foo, 3)
```

On the flip side, now the `TangibleThings` controller test isn't indented like it usually would be, but since our developers will mainly be looking at their Super Scaffolded tests, I prefer having it this way.

## `BlockManipulator#shift_block`
Using `BlockManipulator#shift_block` to shift the content after the file has been generated is another option in case we want to keep the indentation for the `TangibleThings` test as is.

### DIFF
--- a/test/controllers/api/v1/scaffolding/completely_concrete/tangible_things_controller_test.rb
+++ b/test/controllers/api/v1/scaffolding/completely_concrete/tangible_things_controller_test.rb
@@ -2,150 +2,150 @@ require "controllers/api/v1/test"
 
 class Api::V1::Scaffolding::CompletelyConcrete::TangibleThingsControllerTest < Api::Test
   unless scaffolding_things_disabled? # 🚅 skip when scaffolding.
-    def setup
-      # See `test/controllers/api/test.rb` for common set up for API tests.
-      super
+  def setup
+    # See `test/controllers/api/test.rb` for common set up for API tests.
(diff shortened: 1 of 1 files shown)


gazayas:
Oof, didn't even think about Standard Ruby, maybe we will have to use `BlockManipulator#shift_block`

gazayas:
Closing in favor of https://github.com/bullet-train-co/bullet_train-core/pull/78

PR #632, opened 2023-01-23
Add basic healthcheck endpoint

ipmb:
Helpful for hosting environments which require/allow a healthcheck on the app. Some examples:

* https://docs.aws.amazon.com/apprunner/latest/dg/manage-configure-healthcheck.html
* https://cloud.google.com/run/docs/configuring/healthchecks
* https://render.com/docs/deploys#health-checks


andrewculver:
@ipmb Hey Peter, thanks for putting https://github.com/rails/rails/pull/46936 on my radar. We'll wait for it!

PR #665, opened 2023-02-20
OpenAPI system test is flakey

### LINKED ISSUE #664: OpenAPI test is flakey
https://github.com/bullet-train-co/bullet_train/blob/82f42b75c7e58537658755d641c3e14261322586/test/system/open_api_v3_test.rb#L1-L9

I've had issues with this test being flakey for reasons that have been difficult to track down. Generally, it fails because the redocly tool makes an API call to the application stack and sometimes that request fails.  

In thinking about how to make this more robust I ended up concluding that this test really isn't a system test anyway as there is no reason to involve a browser in testing an API endpoint response. I  ended up refactoring it as a controller test which seems to be less error-prone.

I'll open a PR.

olbrich:
- rewrite as a controller test

Fixes #664

### DIFF
--- a/test/controllers/api/open_api_controller_test.rb
+++ b/test/controllers/api/open_api_controller_test.rb
@@ -1,0 +1,15 @@
+require "controllers/api/v1/test"
+require "tempfile"
+
+class Api::OpenApiControllerTest < Api::Test
+  test "OpenAPI document is valid" do
+    get api_path(version: "v1")
+    output = nil
+    Tempfile.create do |f|
+      f.write(response.body)
+      f.rewind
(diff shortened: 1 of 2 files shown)


andrewculver:
@olbrich Thanks! 🙏

newstler:
This test doesn't show warnings. Should we return them to STDOUT? @olbrich @andrewculver

olbrich:
@newstler I think the warnings will end up in the output, so maybe can revise it to fail on a warning.

newstler:
@olbrich Sounds reasonable. I think I can do it when I finish #734

newstler:
Just realised: there are bugs in Redocly linter producing [false warnings](https://github.com/Redocly/redocly-cli/issues/829#issuecomment-1528849724).

PR #674, opened 2023-03-01
Enroll in Showcase's testing

kaspth:
This is just to get things going, we can wait on >= 1.2.11 of bullet_train, so we can remove showcase-rails from the app Gemfile here, I just need it to see the tests run right now.

We bundle `rouge` for out-of-the-box syntax highlighting in Showcase too.


kaspth:
Gonna tackle things another time, or differently, we'll see.

PR #700, opened 2023-03-15
Update .gitignore to not ignore .tool-versions

olbrich:
We don't want to ignore `.tool-versions`.  It's used by `asdf` to determine which version to use for ruby, nodejs, and others.  Ignoring it would be like ignoring `.ruby-version`.


andrewculver:
@olbrich Thanks, Kevin! 🙏

PR #705, opened 2023-03-22
Add initializer for Bullet Train configuration

gazayas:
We have a `BulletTrain::Configuration` class, but we don't implement it in any of our initializers, so I added one here.

Joint PR:
- https://github.com/bullet-train-co/bullet_train-core/pull/208


jagthedrummer (approved):
This looks good once we can get the dependent PR passing CI and merged.

PR #706, opened 2023-03-23
Replace sidebar with navbar

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/209

Since there were changes to the `--sidebar` flag in the joint PR, I updated the tests here so everything passes.


jagthedrummer:
@gazayas I've been looking at this and the joint PR and I think they're both good to merge. My only reservation is that this introduces a change that is a breaking change when the Super Scaffolding tests are run against the published gems.

I think we'll want to coordinate merging of this and the joint PR with @andrewculver so that we do it at a time when he can cut a new release of the `core` gems, and then we can bump the version numbers in the starter repo to get the `Release:` tests working again.

Issue #703, opened 2023-03-20
Conditionally run github merge action


olbrich:
https://github.com/bullet-train-co/bullet_train/blob/a17e073d4a09786b2c7b188e7cc83d4c3816a455/.github/workflows/main.yml#L3-L6

This action seems to be convenient for the maintainers, but may not be at all useful for downstream projects.  Perhaps it needs to be configured to run conditionally when the repo matches `bullet-train-co/bullet_train`.

https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution

I'm not sure if downstream projects would want to use it, but maybe they could make a copy without the condition in a different workflow file to avoid merge conflicts.

Related to #699

jagthedrummer:
We removed the file in question recently.

PR #717, opened 2023-04-03
Cleanup GitHub action when creating new application

### LINKED ISSUE #714: Remove GitHub Action workflow on `bin/setup`
On the Bullet Train application I'm working on the GitHub Action which merges `main` to `edge` and pushes it to GitHub runs and then fails (because there's no `edge` branch), but if my understanding is correct, this is only necessary for the starter repository itself and not actual Bullet Train apps (we don't automatically make an `edge` branch when creating a new app so git has nothing to push to, and I think our developers will only need to run their tests against the gems which have been officially released).

I think we can easily delete the file in the `bin/setup` script.

gazayas:
Closes #714 and #703.

Besides changing the CircleCI config to run tests without Knapsack Pro, all tests will now pass when spinning up a new app:

App Without GitHub Action Workflow

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -113,6 +113,9 @@ else
   end
 end
 
+File.open('.github/workflows/main.yml', 'w') do |f|
+  f.write("# Use GitHub Actions to automate development workflows for your application: https://docs.github.com/en/actions")
+end
 
 puts "Next, let's push your application to GitHub."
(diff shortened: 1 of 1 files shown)


andrewculver:
Also solved by https://github.com/bullet-train-co/bullet_train/pull/721.

PR #720, opened 2023-04-04
[Action Required] Prevent merge conflicts for `device_test`

kaspth:
We're adding a new `device_test` helper that simplifies the structure of system tests.

In a following PR (#964) we'll update all of the system tests in the starter repo, but this might cause nasty merge conflicts for downstream developers. To reduce the chances of merge conflicts we're shipping a helper script that will make the expected changes to your system tests.

To use the helper to modify your tests run this in a console (after updating to v1.4.11):

```
bin/update/system_tests/use_device_test
```

Then you should:

* Run the tests to make sure they still pass
* Commit the updated test files
* Update to v1.5.0

### About this change

We've introduced a new `device_test` helper that wraps up some of the implementation
details about running system tests on a variety of devices.

It simplifies the way that we write system tests like this:

```diff
-  @@test_devices.each do |device_name, display_details|
-    test "user can so something on a #{device_name}" do
-      resize_for(display_details)
-      # actual tests here
-    end
-  end
+  device_test "user can do something" do
+    # actual test code
+  end
```

The specific changes are:

* Remove the enclosing @@test_devices.each do block
* Remove one level of indentation from test inside those blocks
* Remove the resize_for(display_details) from tests in those blocks
* Use `device_test` helper to handle running the test on different devices

### Original PR description

Currently, to get device specific tests users have to wrap their tests in `@@test_devices.each`, which kinda spills out implementation details to apps.

Instead, we can add a `device_test` to abstract those details, which means apps get less indentation and less details they need to care about.

We're also introducing a `bin/updates/system_tests/use_device_test` script to help update app system tests. Apps must run this as described in the commit message. In https://github.com/bullet-train-co/bullet_train/pull/964 we'll update Bullet Train itself via the script.

In a later PR, methods like `sign_out_for(display_details)` now being able to just be `sign_out`, as well as the other helpers doing the same.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -89,13 +89,47 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     key = ENV["TEST_DEVICE"].to_sym
     if @@test_devices.key?(key)
       puts "Running tests with the `#{ENV["TEST_DEVICE"]}` device profile specifically.".green
-      @@test_devices = {key => @@test_devices[key]}
+      @@test_devices = @@test_devices.slice(key)
     else
       puts "⚠️ `#{ENV["TEST_DEVICE"]}` isn't a valid device profile in `test/test_helper.rb`, so we'll just run *all* device profiles.".yellow
     end
   end
 
+  def device_name
+    @device_name ||= @@test_devices.each_key.find { name.include?(_1.to_s) } || raise("unknown test device, you probably want to use `device_test` to generate this test")
+  end
+
+  def display_details
+    @@test_devices[device_name]
+  end
+
+  # Generate a device specific test for each device in `@@test_devices`.
+  #
+  # Automatically resizes the display for the test too.
+  #
+  #   device_test "system test" do
+  #     p device_name
+  #     p display_details
+  #   end
+  def self.device_test(name, &block)
+    @@test_devices.each_key do |device_name|
+      test "#{name} on a #{device_name}" do
+        resize_display # TODO: Figure out why the browser window opens if this is done in `setup`.
+        instance_eval(&block)
+      end
+    end
+  end
+
   def resize_for(display_details)
+    ActiveSupport::Deprecation.warn <<~END_OF_MESSAGE
+      `resize_for` is deprecated.
+      Please run the following command to update your tests:
+      ./bin/updates/system_tests/use_device_test
+    END_OF_MESSAGE
+    resize_display(display_details)
+  end
+
+  def resize_display(display_details = self.display_details)
     if use_cuprite?
       page.driver.resize(*calculate_resolution(display_details))
     else
@@ -103,20 +137,23 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     end
   end
 
-  def within_team_menu_for(display_details)
+  def within_team_menu(display_details = self.display_details)
     first("#team").hover
     yield
   end
+  alias_method :within_team_menu_for, :within_team_menu
 
-  def within_user_menu_for(display_details)
+  def within_user_menu(display_details = self.display_details)
     find("#user").hover
     yield
   end
+  alias_method :within_user_menu_for, :within_user_menu
 
-  def within_developers_menu_for(display_details)
+  def within_developers_menu(display_details = self.display_details)
     find("#developers").hover
     yield
   end
+  alias_method :within_developers_menu_for, :within_developers_menu
 
   def within_membership_row(membership)
     within "tr[data-id='#{membership.id}']" do
@@ -141,7 +178,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   end
 
   # sign out.
-  def sign_out_for(display_details)
+  def sign_out(display_details = self.display_details)
     if display_details[:mobile]
       open_mobile_menu
     else
@@ -152,8 +189,9 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     # make sure we're actually signed out.
     assert_text "Signed out successfully"
   end
+  alias_method :sign_out_for, :sign_out
 
-  def new_session_page_for(display_details)
+  def new_session_page(display_details = self.display_details)
     # We actually want to go straight to the user session and
     # not the home page in case the home page doesn't have new session logic.
     visit new_user_session_path
@@ -162,15 +200,17 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     # otherwise our tests will immediately start trying to match things before the page even loads.
     assert_text("Sign In")
   end
+  alias_method :new_session_page_for, :new_session_page
 
-  def sign_in_from_homepage_for(display_details)
-    puts "sign_in_from_homepage_for is deprecated".red
-    puts "  please switch to: new_session_page_for".red
+  def sign_in_from_homepage(display_details = self.display_details)
+    puts "#{__callee__} is deprecated".red
+    puts "  please switch to: new_session_page".red
     puts "  called from #{caller(1..1).first}".red
     new_session_page_for(display_details)
   end
+  alias_method :sign_in_from_homepage_for, :sign_in_from_homepage
 
-  def new_registration_page_for(display_details)
+  def new_registration_page(display_details = self.display_details)
     # TODO the tailwind port of bullet train doesn't currently support a homepage.
     visit new_user_registration_path
 
@@ -178,27 +218,31 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     # otherwise our tests will immediately start trying to match things before the page even loads.
     assert_text("Create Your Account")
   end
+  alias_method :new_registration_page_for, :new_registration_page
 
-  def sign_up_from_homepage_for(display_details)
-    puts "sign_up_from_homepage_for is deprecated".red
-    puts "  please switch to: new_registration_page_for".red
+  def sign_up_from_homepage(display_details = self.display_details)
+    puts "#{__callee__} is deprecated".red
+    puts "  please switch to: new_registration_page".red
     puts "  called from #{caller(1..1).first}".red
     new_registration_page_for(display_details)
   end
+  alias_method :sign_up_from_homepage_for, :sign_up_from_homepage
 
-  def within_homepage_navigation_for(display_details)
+  def within_homepage_navigation(display_details = self.display_details)
     if display_details[:mobile]
       open_mobile_menu
     end
     yield
   end
+  alias_method :within_homepage_navigation_for, :within_homepage_navigation
 
-  def within_primary_menu_for(display_details)
+  def within_primary_menu(display_details = self.display_details)
     open_mobile_menu if display_details[:mobile]
     within ".menu" do
       yield
     end
   end
+  alias_method :within_primary_menu_for, :within_primary_menu
 
   def be_invited_to_sign_up
     # if the application is configured to only allow invitation-only sign-ups, visit the invitation url.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,6 +125,16 @@ jobs:
           environment:
             KNAPSACK_PRO_CI_NODE_TOTAL: 2
 
+      # Run any update scripts + retest
+      - run: bin/updates/system_tests/use_device_test
+      - run:
+          name: Rerun system tests after update
+          command: |
+            export RAILS_ENV=test
+            KNAPSACK_TEST_FILE_PATTERN="test/system/**/*.rb" bundle exec rails "knapsack_pro:queue:minitest[--verbose]"
+          environment:
+            KNAPSACK_PRO_CI_NODE_TOTAL: 16
+
       # If you don't want to use Knapsack Pro, then use this configuration:
       #
       # - run:
--- a/bin/updates/system_tests/use_device_test
+++ b/bin/updates/system_tests/use_device_test
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "pathname"
+
+Pathname.glob("test/system/**/*.rb") do |path|
+  lines = path.readlines
+  if (devices_line = lines.grep(/@@test_devices\.each do/).first)
+    indentation = devices_line.match(/^\s+/).to_s
+    end_line_index = lines.rindex "#{indentation}end\n"
+
+    # Act on all the lines within the `@@test_devices.each` block
+    lines.slice(lines.index(devices_line).succ...end_line_index).each do |line|
+      # Replace `test` with `device_test`
+      if line.sub!(/(\s+)test/, "\\1device_test")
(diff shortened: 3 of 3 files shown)


test/system/account_management_test.rb
```diff
@@ -1,53 +1,46 @@
...
+    fill_in "Last Name", with: "McTesterson"
+    fill_in "Your Team Name", with: "The Testing Team"
+    page.select "Brisbane", from: "Your Time Zone"
+    click_on "Next"
 
-      fill_in "Email", with: "andrew.culver.new@gmail.com"
-      fill_in "First Name", with: "Testy.new"
-      fill_in "Last Name", with: "McTesterson.new"
-      page.select "Tokyo", from: "Your Time Zone"
+    assert_text "The Testing Team’s Dashboard"
```
kaspth, line 27: In the two sample cases I've replaced `assert page.has_content?(…)` with the equivalent `assert_text` since the latter is easier to read, and helps demonstrate better testing practices to Bullet Train apps.

test/system/account_management_test.rb
```diff
@@ -1,53 +1,46 @@
...
 
-      assert page.find("#user_email").value == "andrew.culver.new@gmail.com"
-      assert page.find("#user_first_name").value == "Testy.new"
-      assert page.find("#user_last_name").value == "McTesterson.new"
-      assert page.find("#user_time_zone").value == "Tokyo"
-    end
+    assert_field "Email", with: "andrew.culver.new@gmail.com"
+    assert_field "First Name", with: "Testy.new"
+    assert_field "Last Name", with: "McTesterson.new"
+    assert_field "Your Time Zone", with: "Tokyo"
```
kaspth, line 44: Same deal as the `assert_text` change, this now pretty much becomes a 1-1 of the `fill_in`/`select` update above.

test/application_system_test_case.rb
```diff
@@ -89,41 +89,70 @@ def teardown
...
+  # Automatically resizes the display for the test too.
+  #
+  #   device_test "system test" do
+  #     p device_name
+  #     p display_details
+  #   end
+  def self.device_test(name, &block)
+    @@test_devices.each_key do |device_name|
+      test "#{name} on a #{device_name}" do
+        resize_display # TODO: Figure out why the browser window opens if this is done in `setup`.
```
kaspth, line 117: It would be nice if `resize_display` could just become part of `setup` but then the browser window opens, and stays open, for some reason.

andrewculver: Yeah, I had this same thought recently! Didn't realize it would cause an issue though...

andrewculver:
100% here for this change, but the indentation change on the system test files is going to kill people when they're merging from the starter repo... I don't have a good solution for this at the moment, other than to put a useless `begin ... end` block around them for now to maintain the indentation level, or create a script that will automatically reindent people's tests (removing the `@@` lines) _before_ they merge this update. 😬 Will keep noodling on this.

jagthedrummer:
I agree that this is a great change and that the indentation change is going to make for a really messy update. I'll also try to think about solutions for that problem. 🤔

kaspth:
Here's what I'm thinking for the update sequence:

1. I can remove the changes to the specific test file here, then users have got the `device_test` helper and they can update manually.
2. Then I can add a script to rewrite and reindent the tests that users can call when they want to. 3. We could use the script to update the starter repo tests and give users a month to run the script on their end. Then any resulting diff should hopefully be easy to sort through.

@andrewculver @jagthedrummer how does this sound to you?

I'm not sure where we'd announce these changes to users, just in the commit message? I could add a `Action:` prefix to indicate that we require an action from them. Sound good?

jagthedrummer:
@kaspth That sounds like a reasonable approach to me. Figuring out where/how to announce it is a good question. Maybe we add a temporary section to the upgrade guide that mentions it?

kaspth:
@jagthedrummer what's the upgrading guide, this one? https://bullettrain.co/docs/upgrades — that seems to target the surface level, but ideally we'd want something more like a changelog.

jagthedrummer:
@kaspth Yep, that's the upgrade guide I was thinking of. I agree that a proper changelog would be a good thing to introduce.

kaspth:
Added a script, but my system tests fail locally, I have to go, but I'll come back and run them on CI later.

jagthedrummer:
> Now that you've setup GitHub releases for Bullet Train, how do I use that to communicate what apps need to do?

I'm not 100% sure on the answer to that question, but here's my first stab at thinking through it.

Our ultimate goal isn't so much to land _this_ PR (which is definitely a good and helpful PR, don't get me wrong). The ultimate goal is to be able to land a _different_ PR that will consist of updating a bunch of tests to be simpler.

We want to take test files that look like this:

```ruby
  @@test_devices.each do |device_name, display_details|
    test "user can so something on a #{device_name}" do
      resize_for(display_details)
      # actual tests here
    end
  end
```

And make them look like this:

```ruby
  test "user can do something on a #{device_name}" do
    # actual tests here
  end
```

Specifically we want to:
* Remove the enclosing `@@test_devices.each do` block
* Remove one level of indentation from test inside those blocks
* Remove the `resize_for(display_details)` from tests in those blocks

(Do we also want to be able to remove `on a #{device_name}` from the description?)

So, I think that _this_ PR is safe to merge at any time because it won't force a change on anyone.

I think we probably want to merge it and have it "in circulation" for a while before trying to land the follow-on PR, and I think _that_ PR will be the one that needs some good messaging.

The other PR that makes all of those simplifications has the possibility of causing hairy merge conflicts. So we're pre-shipping a script that will help people side-step those conflicts. And _that's_ the PR that will require some finesse and good communication.

For that other PR I think we'll want to:

* Identify the target version of BT that we want it to land in. Possibly a major version bump? Let's call it `2.0.0` for now, just for illustration.
* Update the upgrade docs to mention that there's a big change in going from `1.3.X` to `2.0.0`. And that the way to navigate the change is to:
  * Upgrade to the last-available `1.3.x`.
  * Make sure everything is good and tests are passing
  * Run `bin/updates/system_tests/use_device_test`
  * Make sure everything is good and tests are passing
  * Commit the changes made by `bin/updates/system_tests/use_device_test`
  * Upgrade to `2.0.0`
  * Hopefully not have any hairy merge conflicts due to indentation changes

As to how to use our new release process specifically to communicate this stuff, I'm not entirely sure. Release notes are auto-generated based on what's been merged, but we can always go in and edit them to add additional info. [(Example release notes for `1.3.22`.)](https://github.com/bullet-train-co/bullet_train/releases/tag/v1.3.22)

It might be good to have some sort of warning in the title of the PR that's going to update those tests. Something like `[Potentially Breaking Change] Update system tests to remove 'test_devices.each'`. Just something to maybe get someone's attention if they're skimming the PRs that went into that release.

What do you think, @kaspth? Does this sound generally like the right thing to do?

kaspth:
> Our ultimate goal isn't so much to land this PR (which is definitely a good and helpful PR, don't get me wrong). The ultimate goal is to be able to land a different PR that will consist of updating a bunch of tests to be simpler.

Yes, we're shipping an update script in here that Bullet Train apps can use so they can independently have their tests updated first. Then they won't see a ton of merge conflicts after we've run the update script and they pull commits.

> And make them look like this:

We need to use `device_test` too. The script I've added already converts those examples you listed, and changes the code to:

```diff
-  @@test_devices.each do |device_name, display_details|
-    test "user can so something on a #{device_name}" do
-      resize_for(display_details)
-      # actual tests here
-    end
-  end
+  device_test "user can do something" do
+    # actual test code
+  end
```

So yes, I'm also removing `on a #{device_name}` since `device_test` inserts that.

> So, I think that this PR is safe to merge at any time because it won't force a change on anyone.
>
> I think we probably want to merge it and have it "in circulation" for a while before trying to land the follow-on PR, and I think that PR will be the one that needs some good messaging.
>
> The other PR that makes all of those simplifications has the possibility of causing hairy merge conflicts. So we're pre-shipping a script that will help people side-step those conflicts. And that's the PR that will require some finesse and good communication.

I think this PR needs the messaging in it too, since it's easier to communicate alongside the code. E.g. what we expect of people and what. I'm still in favor of prefixing the merge commit with `[Action Required]`.

Ok, I think I have an idea of what to do based on your release process idea here. Let me push some stuff and then I'll explain.

kaspth:
@jagthedrummer ok, I've prepped the v2.0.0 milestone PR over here https://github.com/bullet-train-co/bullet_train/pull/964, it'll update to target `main` after we merge this. We can always change the milestone too.

Once I see CI passing with rerunning tests after running the script, I'll squash the commits and try to give some more info to apps in the commit message. I'll also link up the #964 in there so apps know what it looks like when they run the script.

kaspth:
Ok, looks like something is going on with CI. I'll take a look.

kaspth:
@jagthedrummer how does the commit message in [4f1b44c](https://github.com/bullet-train-co/bullet_train/pull/720/commits/4f1b44c717d3289fb09578b60a8e462287a6bd93), explaining what apps need to do, sound to you?

jagthedrummer:
@kaspth in general I think that description is headed in the right direction.

But I really don't think that we want to encourage people to run the script until they're ready to upgrade to whatever version has the modified versions of the tests in the starter repo.

Like, imagine this scenario:

* We merge this PR and release it
* A dev updates and runs the script, fixing their local tests
* Before the big `2.0.0` release we need to modify `test/system/account_test.rb`
* The dev updates to that new version

I _think_ that they'd potentially end up with a merge conflict in `test/system/account_test.rb` (depending on the nature of the change we had to make). Or at least the changes would seem weird because indentation levels would be off.

If I'm right about that, then I think we don't want to encourage people to run that script until they're doing it as a pre-upgrade step for the `2.0.0` release. And if we mention it at all before then, we probably want to encourage people _not_ to run it. And we definitely don't want to tag this PR with `[Action Required]`. (Again, that's if I'm right about the potential for merge conflicts coming from the other direction.)

jagthedrummer:
I just realized that I don't think there's really any reason to have this out in the wild for a while. Previously we'd discussed doing that, but it was before we'd started explicitly versioning the starter repo.

But since we have started versioning it we could just sit on this PR until we're almost ready for `2.0.0` (or whatever the version is going to be). Then we add it in the release immediately before `2.0.0` (let's call it `1.99.0`) along with the `[Action Required]` prefix. And we also add a new section/page on the `/docs` site that talks about how to get over the hump to `2.0.0`. That would explicitly say to update to `1.99.0`, then run the script, then go to `2.0.0`.

kaspth:
> I just realized that I don't think there's really any reason to have this out in the wild for a while. Previously we'd discussed doing that, but it was before we'd started explicitly versioning the starter repo.

I don't know when/if `2.0.0` is happening, and I don't think we should sit on this for potentially months. There's absolutely a concern about the upgrade burden involved and how that happens, but I don't know if we have to be this cautious?

Since we're on `1.3.x`, I'd rather we decided on `1.4.0` or `1.5.0`, that would follow a shifted SemVer akin to Rails' 6.0, 6.1 etc. 

I'll see if I can figure out how to verify if there's going to be merge conflicts or not for your earlier comment.

jagthedrummer:
@kaspth, yeah, sorry, I wasn't meaning to say that we should sit on this for months. I was just using `2.0.0` as a convenient shorthand to discuss the general concept. I figured that was better than trying to be super precise about exactly _which_ version we expect this to land in. I think `1.4.0` is probably OK.

jagthedrummer:
@kaspth, I just released `1.4.0` as the first release where we explicitly pin the version number of `core` gems in the `Gemfile` of the starter repo. So I'm thinking that we'll target `1.5.0` (maybe next week?) as when we'll release #964.

To update the list from above, and make it a checklist, I think the process for landing this and the associated PR are:

- [ ] Get #964 passing tests, which hopefully won't uncover any new complications. (Btw, the diff for #964 seems a little weird to me because it's showing that it wants to put things into this branch that are already in this branch. I would expect it to show only the changes in `test/system/**`.)
- [x] I'll work on a fix for https://github.com/bullet-train-co/bullet_train-core/issues/470 which will give us a good place to add a call-out about the `1.4.x` => `1.5.0` transition.
- [ ] We write up the transition process somewhere in the docs and link it in the call-out mentioned above. (I can also handle this.)
- [ ] We give people in Discord a heads up that the `1.5.0` release is coming. Are there other places we should also do this? Maybe ask Andrew to mention it on his various social accounts?
- [ ] We merge this PR (with the `[Action Required]` alert in the PR title)
- [ ] We release version `1.4.x` in both `core` and starter
- [ ] We merge #964
- [ ] We release version `1.5.0` in both `core` and starter

kaspth:
@jagthedrummer this sounds good to me! I'll work on #964 (I'm also confused why it's not just the `test/system/**` changes in there).

> We give people in Discord a heads up that the 1.5.0 release is coming. Are there other places we should also do this? Maybe ask Andrew to mention it on his various social accounts?

Sure yeah, we could ask Andrew about this on Discord.

jagthedrummer:
@kaspth, excellent! I'm going to have a short day today and then will be out for a long weekend. I want to have plenty of time to get the merge & release for these right, so I'm going to leave them until next week. They'll be on the top of my list when I'm back.

kaspth:
@jagthedrummer perfect, sounds good! Enjoy your long weekend!

jagthedrummer:
@kaspth, I just had one more thought on this. In the case where someone merges this PR from upstream but _does not_ run the update script, we should probably display a deprecation messages in the console or raise an error or something.

Maybe by having `resize_for` continue to be a "real" method that displays the deprecation message before calling `resize_display`?

Something like:

```ruby
def resize_for(display_details)
  puts "--------------------------------------------------------".red
  puts "resize_for is deprecated.".red
  puts "please run the following command to update your tests:".red
  puts "./bin/updates/system_tests/use_device_test".red
  puts "--------------------------------------------------------".red
  resize_display(display_details)
end
```

test/application_system_test_case.rb
```diff
@@ -89,41 +89,78 @@ def teardown
...
+  end
+
   def resize_for(display_details)
+    ActiveSupport::Deprecation.warn <<~end_of_message
+      `resize_for` is deprecated.
+      Please run the following command to update your tests:
+      ./bin/updates/system_tests/use_device_test
+    end_of_message
+    resize_display(display_details)
+  end
```
kaspth, line 130: @jagthedrummer I added the deprecation warning like you mentioned, but I opted for Rails' built-in warning mechanism.

gazayas:
I'm late to the conversation and I know this one is on the verge of being merged, just wanted to mention that we also have the [Scaffolding::BlockManipulator](https://github.com/bullet-train-co/bullet_train-core/blob/5a917dde2f1da7183ccb85685257ce5374b3c9ca/bullet_train-super_scaffolding/lib/scaffolding/block_manipulator.rb) in case we wanted to write any scripts that update block indentation, etc. in the future.

Looking forward to using this new `device_test` setup in my own app!

kaspth:
eyy nice! 🎉

@gazayas good point, I'll try to keep a note for that if we need it in the future 😄

PR #721, opened 2023-04-06
Remove default github workflow file in configure before pushing to github

cnorm35:
Resolves bug reported in the BT Discord channel about failing GitHub jobs

- Removes the included GitHub workflow config to merge main into edge in `bin/configure`
- Guards against throwing a failed job error after pushing to GitHub the first time.

Wasn't sure if it made sense to include a prompt before removing so left that out for the time being.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -113,6 +113,11 @@ else
   end
 end
 
+if File.exist?('.github/workflows/main.yml')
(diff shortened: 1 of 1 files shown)


andrewculver:
@cnorm35 Thanks Cody! 🙏

PR #722, opened 2023-04-07
Fix: uninitialized constant FactoryBot::Syntax on Heroku

newstler:


iwan:
Got the same error on Render

newstler:
> 

On this branch?

olbrich:
This probably shouldn't even be in an initializer, but should be done in `test_helper.rb`. If there's a good reason for this elsewhere it should be documented.

test/test_helper.rb
```diff
@@ -27,3 +27,9 @@
 
   # Add more helper methods to be used by all tests here...
 end
+
+if defined?(FactoryBot)
```
olbrich, line 31: I don't think you need the conditional anymore since FactoryBot is a dependency in the test group.

newstler:
@olbrich Good point! Moved code to `test_helper.rb` and updated PR.

PR #726, opened 2023-04-10
Include locales with bullet_train-internationalization

gazayas:
https://github.com/bullet-train-co/bullet_train-core/pull/239


jagthedrummer (approved):
I think this is good to go if we can get the joint PR de-conflicted and passing tests.

gazayas:
Closing in favor of #855.

PR #660, opened 2023-02-17
Add system tests for `number_field`

gazayas:
Pretty straightforward, I just added tests for https://github.com/bullet-train-co/bullet_train-core/pull/113.

PR #728, opened 2023-04-15
Enable password field tests, move Super Scaffolding file tests

gazayas:
Addresses a couple of TODOs in the Super Scaffolding test setup script.

As far as `TestFile` goes, we already had a test for it in the partial test, but we can go with this PR if it belongs in `super_scaffolding_test.rb` instead.


andrewculver:
@gazayas <a href="https://imgflip.com/i/7jz0wz"></a><div><a href="https://imgflip.com/memegenerator">from Imgflip Meme Generator</a></div>

gazayas:
Meme saved! 👆

Hadn't checked my fork in a while, 22 commits behind will do it 😅️
22 commits behind

jagthedrummer:
@gazayas Looks like another conflict has cropped up. If you can resolve it I'll merge this ASAP to avoid any more conflicts.

gazayas:
It looks like some of these tests were covered in #86 and #660, so there's not much left here after merging. We do get rid of the TODOs though, and add we also add one TODO related to finding a "one" option with Capybara. This one should be good to go!

PR #737, opened 2023-04-20
Invitations API

newstler:
Implements memberships API and adds OpenAPI specs.
Required by https://github.com/bullet-train-co/bullet_train-core/pull/257

Not sure about this line, please correct me if I did it wrong: https://github.com/bullet-train-co/bullet_train/blob/e7133604d0fddc6c797626e3cdf255038ad367e2/app/models/ability.rb#L20


jagthedrummer:
This PR seems like it's probably good to go once we get the joint PR in `core` ready. That one has some merge conflicts and failings tests (which should be the same failures we're seeing in this PR for the `Core:` tests). Just to help me keep track of what's ready for review/merge I'm going to convert this to a draft PR for now.

Issue #734, opened 2023-04-19
Fix OpenAPI warnings for system test


gazayas:
Although the document produced in `open_api_v3_test.rb` is valid, we get a lot of warnings. I sat down for a bit with this one but couldn't quite seem to fix it. I noticed we're using different methods to write out the contents of the yaml file, so it might be worth looking into those to debug:
1. Ruby's [to_yaml](https://github.com/bullet-train-co/bullet_train-core/blob/d1d9be164aa8cfd4a7b06c456f5d0fc2e4529f12/bullet_train-api/lib/bullet_train/api/example_bot.rb#L107)
2. Our [_yaml_output](https://github.com/bullet-train-co/bullet_train-core/blob/d1d9be164aa8cfd4a7b06c456f5d0fc2e4529f12/bullet_train-api/lib/bullet_train/api/example_bot.rb#L140)

newstler:
@gazayas Could you please provide the output with those warnings? That would be very helpful.

gazayas:
Sure thing, it's just the output when running the `open_api_v3_test.rb` system test. Does running the test yield a different result for you?

<details>
  <summary>Test Output</summary>

```
> bundle exec rails test test/system/open_api_v3_test.rb
🌱 Generating global seeds.

Creating a platform application for Zapier. Within the `zapier` directory, run:

  cd zapier
  zapier env:set 1.0.0 BASE_URL=https://gazayas.jp.ngrok.io CLIENT_ID=dyBwxWpm1_51W79AGAxH44-VA3aGrGao9dWgEiHeMQ8 CLIENT_SECRET=mqKtW1wNbP01KTLK5zwLCj60hjNIbWhaOlSn_QPG3YQ
  cd ..

🌱 Generating test environment seeds.
Coverage report generated for Unit Tests to /home/gazayas/work/bt/bullet_train/coverage. 60 / 296 LOC (20.27%) covered.
Run options: --seed 42271

# Running:

Puma starting in single mode...
* Puma version: 6.2.1 (ruby 3.2.1-p31) ("Speaking of Now")
*  Min threads: 5
*  Max threads: 5
*  Environment: test
*          PID: 28111
* Listening on http://127.0.0.1:3001
Use Ctrl-C to stop
yarn exec v1.22.19
No configurations were provided -- using built in recommended configuration by default.

validating http://localhost:3001/api/v1/openapi.yaml...
[...]

PR #749, opened 2023-04-29
Fix: OpenAPI warnings

newstler:
Closes https://github.com/bullet-train-co/bullet_train/issues/734
Required by https://github.com/bullet-train-co/bullet_train-core/pull/270

- Introduces `.redocly.yaml` config to turn off `no-ambiguous-paths` check, as Redocly team contributor [confirmed](https://github.com/Redocly/redocly-cli/issues/829#issuecomment-1531054481) that some paths in our API are considered non ambiguous (`/projects/{project_id}/goals` and `/projects/tags/{id}` for example).
- Rewrote `test/controllers/api/open_api_controller_test.rb` so it fails on any warnings.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -133,7 +133,7 @@ GEM
       bullet_train
       doorkeeper
       factory_bot
-      jbuilder-schema (>= 2.0.0)
+      jbuilder-schema (>= 2.2.0)
       pagy
       pagy_cursor
       rack-cors
@@ -612,8 +612,8 @@ DEPENDENCIES
   active_hash
   aws-sdk-s3
   bootsnap
-  bullet_train!
-  bullet_train-api!
+  bullet_train
+  bullet_train-api
   bullet_train-incoming_webhooks
   bullet_train-integrations
   bullet_train-integrations-stripe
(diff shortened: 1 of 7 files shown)


jagthedrummer:
There are some test failures in the joint PR, which also show up here in the `Core:` tests. Going to convert this to a draft for now.

Gemfile.lock
```diff
@@ -605,7 +601,6 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
-  ruby
```
jagthedrummer, line 608: @newstler, I'm curios about all the removals in `Gemfile.lock`. Was that intentional?

newstler: Most probably not, the only thing that should have changed there is update of `jbuilder-schema` to `2.2.0`. I just restored Gemfile.lock from `main`, updated `jbuilder-schema` again and rebased everything against `main`. Doesn't fail on my local machine

jagthedrummer: 👍 Excellent! The `Release:` failures here are just saying that these changes would break against the versions of the `core` gems that are currently pinned in `Gemfile.lock`. I'll merge the related PR in `core`, then release a new version of the gems, and bump `Gemfile.lock` and then this should be all good. Thanks for your work on it!

jagthedrummer (approved):

PR #753, opened 2023-05-01
Use previous version of Capybara

gazayas:


gazayas:
There was nothing wrong itself with capybara, but it seems that there was a caching error on CircleCI with a previous commit in Capybara

[CircleCI build](https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/1765/workflows/3e881c92-8187-47a3-8baa-5abd497aa64d/jobs/7617):
```
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path 'vendor/bundle'`, and stop using this flag
Bundler 2.4.6 is running, but your lockfile was generated with 2.4.8. Installing Bundler 2.4.8 and restarting using that version.
Fetching gem metadata from https://rubygems.org/.
Fetching bundler 2.4.8
Installing bundler 2.4.8
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path 'vendor/bundle'`, and stop using this flag
Fetching gem metadata from https://rubygems.org/.......
Git error: command `git reset --hard 96a8baab1df6159a4c0790b1b21e875ee946fa7d`
in directory
/home/circleci/project/vendor/bundle/ruby/3.2.0/bundler/gems/capybara-96a8baab1df6
has failed.
fatal: Could not parse object '96a8baab1df6159a4c0790b1b21e875ee946fa7d'.

Exited with code exit status 11
CircleCI received exit code 11
```

After reverting to a previous version here and then reverting back, the error is now gone.

For example, the build for #705 looks like this only after hitting the rerun button:
Screenshot from 2023-05-01 19-57-39

PR #754, opened 2023-05-01
Update yarn packages with v3.5.1, update esbuild to v0.17.18

gazayas:
esbuild needs to be updated because v17 has breaking changes.


gazayas:
Reactivity test is failing, so we'll need to add Cable Ready v5.0.0: https://github.com/bullet-train-co/bullet_train-core/pull/246

gazayas:
Having a hard time upgrading esbuild. Will keep working on this one for a little bit.

gazayas:
Trying to follow along with the docs [here](https://esbuild.github.io/api/#live-reload), can't quite seem to get things working though. There's a tab for JS and I tried to use that code in both `esbuild.config.js` and `app/javascript/application.js`, to no avail.

Anyways, here are some related docs:
[v0.17.0 release](https://github.com/evanw/esbuild/releases/tag/v0.17.0)
[v0.17.17 changelog](https://github.com/evanw/esbuild/blob/main/CHANGELOG.md#01717)

This is the closest I'm getting
```js
// In esbuild.config.js
// Remove `watch` from `require("esbuild").build(...)`

async function serve_with_esbuild() {
  let ctx = await require("esbuild").context({
    entryPoints: {
      ...otherEntrypoints,
      "application": path.join(process.cwd(), "app/javascript/application.js"),
      "intl-tel-input-utils": path.join(process.cwd(), "app/javascript/intl-tel-input-utils.js"),
      ...themeEntrypoints,
    },
    bundle: true,
    outdir: path.join(process.cwd(), "app/assets/builds"),
  })
  
  await ctx.watch()
  
  let { host, port } = await ctx.serve({
    servedir: path.join(process.cwd(), "app/assets/builds")
  })
}

serve_with_esbuild()
```

pascallaliberte:
On my local system, the reactivity test fails even on e08c91ba283d93cc182780cadeb34ac689aec56c, currently `HEAD` on the main branch.

Will keep digging.

pascallaliberte:
Weird.

After going back in time to earlier and earlier commits, still not being able to run that reactivity test, I decided to manually run a `bin/dev` and reproduce the reactivity with two windows. That worked. And then I ran the test. That now worked. Then I pointed my HEAD to this branch, ran bundle then yarn install and re-ran the test locally. It worked.

Re-ran the CI suite from start. Still failing.

pascallaliberte:
@gazayas if you pick it up on your end and you find out what the problem is from my thing above, please go for it. I don't know if I'll be able to go much further this week.

gazayas:
@pascallaliberte I got things to work.

yarn 3.5.x was using Yarn's Plug'n'Play for some reason when running `yarn install`, so it was creating `pnp.cjs` and another file.

However, I reverted back to v1.22.19 and everything worked fine. Then after upgrading to 3.5.1, I ran `yarn install` again and the `pnp.cjs` file wasn't there, but instead `yarnrc.yml` and `.yarn/` were. The tests started working after that.

I'm not familiar enough with Plug'n'Play to say what's going on here, but I think that's the issue.

Will try and see if I can get things running with esbuild v17.

gazayas:
@pascallaliberte Done!!! That was intense :face_with_spiral_eyes:

Pinging you for a review, specifically on the update to esbuild in `esbuild.config.js`. I made sure esbuild watches changes only when the `--watch` flag is passed, so you can run `yarn build` on it's own and it will simply build. However if you run `bin/dev`, you can see that it will run the watch process.

package.json
```diff
@@ -18,9 +17,9 @@
     "@tailwindcss/forms": "^0.5.3",
     "@tailwindcss/typography": "^0.5.8",
     "autoprefixer": "^10.4.14",
-    "cable_ready": "^5.0.0",
+    "cable_ready": "5.0.0-pre8",
```
pascallaliberte, line 20: I think we were at `-pre9` prior to this automated upgrade. Not a blocker I don't think.

pascallaliberte (approved):

package.json
gazayas: No problem, I just use pre8 since that's what we currently have on the main branch

https://github.com/bullet-train-co/bullet_train/blob/dacdcc913cf68d93d685db3897c4180e70804e51/package.json#L21

pascallaliberte: Oh, then all good with pre8

jagthedrummer, line 20: @gazayas, this looks like it would downgrade `cable_ready` from `^5.0.0` to `5.0.0-pre8`. Can you fix that and make sure that everything is still good?

Looks like `esbuild` a few lines down would also lose a `^`, which we may not want.

jagthedrummer (changes requested):
Looks like some dependencies have moved out from under this PR. Once that's resolved I think this should be good to go.

package.json
gazayas: Done!

gazayas:
Fixed the dependencies and merged main, so this one should be good to go.

jagthedrummer:
@gazayas I think something is weird here. I noticed that this PR seems to be including _lots_ of stuff that is already in `main`, and then I noticed that this PR isn't pointed to merge into `main` but into a `depfu` branch. Is that what we want? Would it be cleaner to go straight to `main`?

I guess maybe we could merge this one into the `depfu` branch and then see what the resulting PR for that branch looks like?

What do you think?

gazayas:
@jagthedrummer Oof, that's totally on me, I forgot the PR was pointing to the #750.

Since I merged in `main` I think pointing it there will be the most painless route (thinking about gem version inconsistencies, for example the depfu branch currently has cable_ready v5.0.0.pre9 in `Gemfile.lock`), so I can change that for the time being, but if you think it's better to merge it into #750 I'll change it back and look at the updated depfu branch. Sorry!

jagthedrummer:
No worries, @gazayas! I think this is looking good overall. But I have one final question. Does it seem weird to you that `yarn.lock` nearly doubled in size? It was 2,293 lines long, and after this PR it would be 4,125 lines.

gazayas:
@jagthedrummer
Looking at the differences with the `main` branch, `languageName` and `linkType` are set for each package, and the links for `resolved` are now just a package name without a link:

## `main`
```
"@babel/code-frame@^7.18.6":
  version "7.18.6"
  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.18.6.tgz#3b25d38c89600baa2dcc219edfa88a74eb2c427a"
  integrity sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==
  dependencies:
    "@babel/highlight" "^7.18.6"
```

## This PR
```
"@babel/code-frame@npm:^7.18.6, @babel/code-frame@npm:^7.21.4":
  version: 7.21.4
  resolution: "@babel/code-frame@npm:7.21.4"
  dependencies:
    "@babel/highlight": ^7.18.6
  checksum: e5390e6ec1ac58dcef01d4f18eaf1fd2f1325528661ff6d4a5de8979588b9f5a8e852a54a91b923846f7a5c681b217f0a45c2524eb9560553160cd963b7d592c
  languageName: node
  linkType: hard
```

I've been investing a bit to find out the implication of these differences, but I'm not finding much. The only thing I can really think of is the difference between having `nodeLinker` set in `yarnrc.yml` or not, which will determine if `pnp` files are automatically generated when running `yarn install`. Besides that, I've just tried to use `yarn install` as is without much tweaking.

@pascallaliberte If you don't mind would you taking another look at `yarn.lock` specifically?

## `private: "true"`
Also there seemed to be a change with how `private: "true"` gets processed ([explained here](https://stackoverflow.com/questions/72292076/why-does-yarn-remove-private-true-in-my-package-json)), so I've set the key-value pair back using a boolean instead.

jagthedrummer (approved):

jagthedrummer:
Ah, good find on the format change @gazayas. I should have looked closer. It definitely looks like the format of a single `yarn.lock` entry has changed and is now considerably more verbose. I was worried that we somehow had introduced a bunch of new dependencies, but now I don't think that's it. I think we're good to merge this unless @pascallaliberte has any objections.

pascallaliberte:
> @pascallaliberte If you don't mind could you take another look at `yarn.lock` specifically?

I'm not going to be able to contribute much for the new `yarn.lock` format right now. I don't know what's at stake, but surely other packages/OSS projects have gone through a similar hesitation with the new format and we would have heard some grumblings already.

jagthedrummer:
OK, I think we're good on this one. With the format of `yarn.lock` being more verbose that accounts for the extra lines. And everything that we're specifying directly in `package.json` looks like a good change. I'm gonna merge it and we can always revert if we need to.

gazayas:
Got it, glad this one is merged, thanks to both of you guys for looking into this one in detail.

PR #756, opened 2023-05-01
Update tangible_thing_test.rb

### LINKED ISSUE #755: Tangible things test should start at /sessions/new
https://github.com/bullet-train-co/bullet_train/blob/58ad3c9000fd7b302f9d7fadb7a116f990660273/test/system/tangible_thing_test.rb#L9

This test breaks if you replace the root_path with a different controller. Pointing it at 'sessions/new' will be more reliable.

olbrich:
Change root_path to user_session_path so it will be less brittle. Changing the controller that the root_path points at can break the existing implementation.

Fixes #755

### DIFF
--- a/test/system/invitations_test.rb
+++ b/test/system/invitations_test.rb
@@ -25,7 +25,7 @@ class InvitationDetailsTest < ApplicationSystemTestCase
 
       be_invited_to_sign_up
 
-      visit root_path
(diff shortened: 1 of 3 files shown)


olbrich:
Looks like there are more places this is an issue, expect a few more changes.

PR #758, opened 2023-05-02
YAML expects null, not nil

jagthedrummer:
When a YAML parser encounters `nil` it is parsed as the literal string `'nil'`. But when a YAML parser encounters `null` it is parsed as `nil`.

This PR just changes instances of `nil` to `null` in `config/models/produts.yml`.

### DIFF
--- a/config/models/billing/products.yml
+++ b/config/models/billing/products.yml
@@ -14,7 +14,7 @@ basic:
   limits:
     memberships:
       have:
-        count: nil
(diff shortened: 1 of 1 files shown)


jagthedrummer:
Note that merging this change may cause problems for Pro/usage users until this PR is merged: https://github.com/bullet-train-pro/bullet_train-billing-usage/pull/32

PR #760, opened 2023-05-03
Update packages with yarn 3.5.1

gazayas:
After merging #746 we can use `yarn 3.5.x`, but this generates `.yarn/` which we don't need to push, so I ignored it here.

Besides that, all I did was run `yarn install` with yarn version 3.5.1.


gazayas:
Closing in favor of #754.

PR #776, opened 2023-05-09
Move duplicated code to application system test case

gazayas:
We had some duplicate methods in different places in the system tests, so I placed them all in `test/application_system_test_case.rb`.


jagthedrummer (approved):
🤘

gazayas:
:drum:

Issue #573, opened 2022-12-27
Failing tests in new project


phortx:
There are a lot of tests, that fail after generating a new project. Example:

```
Error:
Api::V1::Platform::AccessTokensControllerTest#test_index:
NameError: uninitialized constant Api::V1::Platform::AccessToken

          access_token = Platform::AccessToken.find(access_token_data["id"])
                                 ^^^^^^^^^^^^^
    test/controllers/api/v1/platform/access_tokens_controller_test.rb:26:in `assert_proper_object_serialization'
    test/controllers/api/v1/platform/access_tokens_controller_test.rb:49:in `block in <class:AccessTokensControllerTest>'
```

Or:

```
rails test test/controllers/account/scaffolding/absolutely_abstract/creative_concepts/collaborators_controller_test.rb:27

E

Error:
Account::Scaffolding::AbsolutelyAbstract::CreativeConcepts::CollaboratorsControllerTest#test_should_update_collaborator:
NoMethodError: undefined method `account_scaffolding_absolutely_abstract_creative_concepts_collaborator_url' for #<Account::Scaffolding::AbsolutelyAbstract::CreativeConcepts::CollaboratorsControllerTest:0x000000010d497628>
    test/controllers/account/scaffolding/absolutely_abstract/creative_concepts/collaborators_controller_test.rb:57:in `block in <class:CollaboratorsControllerTest>'
```

gazayas:
Hey @phortx, could you provide the steps you took to generate a new project? I got tests passing with the following steps:

```
> git clone git@github.com:bullet-train-co/bullet_train.git foo
> cd foo
> bin/configure
> bin/setup
```

siaw23:
Can confirm tests pass fine but only if I add the missing step from the README `bin/setup`.

gazayas:
Yes, we prompt the developer to run `bin/setup` here when running `bin/configure`:

https://github.com/bullet-train-co/bullet_train/blob/bcd7ad6dbc81cae84f59f725e6b04a2e9b2868e5/bin/configure#L212

siaw23:
[...]

PR #786, opened 2023-05-13
Directly check if node is installed in bin/configure

paulhenrich:
While installing and configuring bullet_train, I noticed that it told me I didn't have node installed because it was checking whether I had installed it using homebrew. While I use homebrew for services, for language runtimes I use another package manager.

My thinking is that just directly checking if node is installed and that it's the right version is a somewhat user-friendlier approach, and that it matches how you handle the ruby version.

I moved the yarn check below the node check and now also fail if node isn't installed because I think this is probably the intended behavior if you're actually checking for presence of the runtime.

This is my first PR so please let me know if this type of PR is welcome and if I need to know anything else with respect to contributing.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -90,20 +90,18 @@ end
 check_package("postgresql@14")
 check_package("redis")
 check_package("icu4c")
-check_package("node")
-
-if `yarn -v 2> /dev/null`.length > 0
-  puts "Yarn is installed.".green
-else
-  puts "You don't have Yarn installed. We can't proceed without it. Try `brew install yarn` or see the installation instructions at https://yarnpkg.com/getting-started/install .".red
-  exit
-end
-
 
 required_node = `cat ./.nvmrc`.strip
-actual_node = `node -v`.strip.gsub("v", "")
+actual_node = begin
+                `node -v`.strip.gsub("v", "")
+              rescue
+                :not_found
+              end
 message = "Bullet Train requires Node.js #{required_node} and `node -v` returns #{actual_node}."
-if Gem::Version.new(actual_node) >= Gem::Version.new(required_node)
+if actual_node == :not_found
+  puts "You don't have Node installed. We can't proceed without it. Try `brew install node`.".red
+  exit
+elsif Gem::Version.new(actual_node) >= Gem::Version.new(required_node)
   puts message.green
 else
   puts message.red
@@ -111,6 +109,13 @@ else
   if input.downcase[0] == "n"
     exit
   end
+end
+
+if `yarn -v 2> /dev/null`.length > 0
+  puts "Yarn is installed.".green
+else
+  puts "You don't have Yarn installed. We can't proceed without it. Try `brew install yarn` or see the installation instructions at https://yarnpkg.com/getting-started/install .".red
+  exit
 end
 
 if File.exist?('.github/workflows/main.yml')


paulhenrich:
I did some manual testing and here are the results:

Node not installed at all:
```sh
Bullet Train requires Ruby 3.2.2 and `ruby -v` returns ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin21].
Homebrew is installed.
postgresql@14 is installed via Homebrew.
redis is installed via Homebrew.
icu4c is installed via Homebrew.
You don't have Node installed. We can't proceed without it. Try `brew install node`.
```

Node 18 installed:

```sh
Bullet Train requires Ruby 3.2.2 and `ruby -v` returns ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin21].
Homebrew is installed.
postgresql@14 is installed via Homebrew.
redis is installed via Homebrew.
icu4c is installed via Homebrew.
Bullet Train requires Node.js 19.3.0 and `node -v` returns 18.16.0.
Try proceeding with Node 18.16.0 anyway? [y/n]
n
```

Node 19 installed:

```sh
Bullet Train requires Ruby 3.2.2 and `ruby -v` returns ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin21].
Homebrew is installed.
postgresql@14 is installed via Homebrew.
redis is installed via Homebrew.
icu4c is installed via Homebrew.
Bullet Train requires Node.js 19.3.0 and `node -v` returns 19.3.0.
Yarn is installed.
...
```

gazayas:
> This is my first PR so please let me know if this type of PR is welcome and if I need to know anything else with respect to contributing.

Yes, PRs like this are totally welcome, thank you!

From [line 83](https://github.com/bullet-train-co/bullet_train/blob/616b9552193ddecdb4fa59fb8360fc4aaddcf9ef/bin/configure#L83):
> You don't have Homebrew installed. This isn't necessarily a problem, you might not even be on macOS, but we can't check your dependencies without it.

Since we have this message, our default is to manage everything via homebrew, but developers will definitely be using other operating systems (I mainly use Linux), so I'm totally fine with relying upon the CLI (or whatever we can use to do checks regardless of the package manager or OS) to grab the version.

I think this file needs updating in general for other operating systems, so I think it's a helpful step in the right direction. For example, this is my output when trying to check for `node` with the `not_installed?` method:

```
[5] pry(main)> `brew info #{package} | grep "Not installed"`
sh: 1: brew: not found
=> ""
```

Anyways, the changes look good to me. Will try to keep this one in the back of my mind for the time being.

paulhenrich:
Great, sounds good to me. I figured it might be somewhat common for people to do what I'm doing, which is using homebrew for services and rtx or asdf for languages. One other thing I ran into if you are open to suggestions for improvements was that the message at the end of this script says to run `bin/dev` when I think it should probably suggest running `bin/setup` (or offer to run it for you?). It's in the docs but I forgot to run it for a minute due to the last printed message in the script.

gazayas:
> One other thing I ran into if you are open to suggestions for improvements was that the message at the end of this script says to run bin/dev when I think it should probably suggest running bin/setup (or offer to run it for you?).

I personally agree with this, but there was an issue and a PR in the past that touched upon the subject, so feel free to read about them here:

- https://github.com/bullet-train-co/bullet_train/issues/573
- https://github.com/bullet-train-co/bullet_train/pull/605

PR #787, opened 2023-05-15
Fix system tests for when billing is enabled

### LINKED ISSUE #782: Adding the billing gems breaks the system test suite
After adding the billing gems I'm getting several failures in system tests that look like this:

```
Failure:
InvitationDetailsTest#test_visitors_can_sign-up_and_manage_team_members_with_subscriptions_enabled_on_a_macbook_pro_15_inch [/Users/jgreen/projects/seshy-rails/test/application_system_test_case.rb:287]:
Expected false to be truthy.


rails test test/system/invitations_test.rb:23
```

`application_system_test_case.rb:287` points to the `complete_pricing_page` [method](https://github.com/bullet-train-co/bullet_train/blob/main/test/application_system_test_case.rb#L286-L294), and the screenshot is showing that the user isn't being sent to the pricing page, but is going directly to the next step where they're prompted for their name and timezone.

failures_test_visitors_can_sign-up_and_manage_team_members_with_subscriptions_enabled_on_a_macbook_pro_15_inch

So it seems like the test suite expects the billing gems to do things that the billing gems don't seem to do.

gazayas:
Closes #782.

Since our billing page has changed, we needed to update the test helper that fills out the necessary fields to subscribe to a plan.

# Old
pricing

# New
new_pricing

There are still a bunch of system tests that need to be updated, so I'll mark the PR as ready for review once those are done.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,54 @@ jobs:
       # - store_test_results:
       #     path: test/reports
 
+  'Minitest with HIDE_THINGS':
+    docker:
+      - <<: *ruby_node_browsers_docker_image
+      - <<: *postgres_docker_image
+      - image: circleci/redis
+    executor: ruby/default
+    parallelism: 16
+    steps:
+      - browser-tools/install-browser-tools
+      - checkout
+      - restore_cache: *restore_bundler_cache
+      - restore_cache: *restore_yarn_cache
+
+      # Install dependencies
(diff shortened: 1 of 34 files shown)


jagthedrummer:
@gazayas This looks great! Can you resolve the conflicts?

gazayas:
@jagthedrummer Done!

jagthedrummer (approved):

Issue #788, opened 2023-05-15
First time bin/setup failures


olbrich:
When running `bin/setup` immediately after installing a new ruby version (or probably starting with the project), the setup script will fail because it requires 'colorize', which probably is not installed

https://github.com/bullet-train-co/bullet_train/blob/616b9552193ddecdb4fa59fb8360fc4aaddcf9ef/bin/setup#L1-L3

This forces the user to install `colorize` manually, which is just annoying.

This can be remedied by using [bundler/inline](https://bundler.io/guides/bundler_in_a_single_file_ruby_script.html)

jagthedrummer:
This was fixed in #789.

PR #789, opened 2023-05-15
Use bundler/inline to install dependencies needed by setup script before bundler has run.

olbrich:
Fixes First time bin/setup failures #788

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,6 +1,12 @@
 #!/usr/bin/env ruby
-require "fileutils"
-require "colorize"
+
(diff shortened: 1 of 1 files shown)


olbrich:
This still requires 'bundler' to already be installed, but this is probably a safer assumption and I think bundler now gets installed automatically with the ruby versions used for this project.

jagthedrummer (approved):
This is an excellent change @olbrich. TIL about `bundler/inline`. Thank you!

PR #796, opened 2023-05-18
feature: Add Avo admin panel

adrianthedev:
This PR adds Avo to Bullet Train as the default admin panel.

I configured four resources: `User`, `Team`, `Membership`, and `Invitation` with a few sensible starter fields.

If `current_user.is_developer?` they can access the admin at `/admin/avo`.

Learn more about Avo here: <https://avohq.io/>

Supersedes https://github.com/bullet-train-co/bullet_train/pull/794


jagthedrummer:
@adrianthedev would you mind rebasing this against the current `main` branch? That might clear up the CI failures you're seeing.

adrianthedev:
CleanShot 2023-08-11 at 22 55 35@2x

@jagthedrummer I'm not sure if I should run `bundle update capybara` on this PR.

```
Downloading capybara-3.39.0 revealed dependencies not in the API or the lockfile
(nokogiri (~> 1.8)).
Running `bundle update capybara` should fix the problem.

Exited with code exit status 34
```

adrianthedev:
Ok. I did a proper rebase...

jagthedrummer:
Hey @adrianthedev, I just pulled this branch down and am poking around at the `/avo` area. This is my first encounter with Avo, so I apologize if this is a dumb question/concern.

I'm a little concerned about how it would perform for busy apps. I'm looking at the "Edit Membership" screen for instance, and it has `select` elements for both `team` and `user` and it looks like it pre-loads _every_ team and user into those lists. Do you know how this performs against large, active apps? Like if there are 10k users and 2k teams, it seem like that screen would take forever to load and would probably cause system degradation while it was loading.

adrianthedev:
No worries @jagthedrummer. No dumb questions. I love all the feedback.

That's correct. That wouldn't perform well.
The fix there is to add `searchable: true` to that field. That will not load any records but bring up a search interface to find those records.

I probably should add it though.

jagthedrummer:
Ah-ha! I'm glad to hear that there's an easy fix. Other than that I think it's worth mentioning that the search field doesn't work, and I can see errors coming through the backend.

adrianthedev:
This is something that I'd like to go over with you. So we use `ransack` for search. It's in the `Gemfile`, but commented out.
If we enable ransack we should also add `ransackable_attributes` to the models in order to make them searchable. 

What do you advise to do? Enable ransack in the `Gemfile` and have the search working or disable the search in Avo (comment out the code that configures the search) and let the users enable it when they need it?

jagthedrummer:
🤔 Hmm... lemme think about that a bit. Neither option just jumps out as being the _obviously_ correct one.

adrianthedev:
Yes. The `searchable` option is using the search and the users' configured `search_query`.

See the Course on Course Link in the avodemo app [here](https://main.avodemo.com/avo/resources/course_links/3/edit) and the corresponding [source code](https://github.com/avo-hq/main.avodemo.com/blob/main/app/avo/resources/course_link.rb#L36).

jagthedrummer:
OK, @adrianthedev I think that info makes an obvious choice stand out.

I think we _definitely don't_ want to ship an implementation of Avo that includes pre-loaded `select` lists. So that means we need to include `ransack` anyway, which means that we may as well enable the default search and make everything searchable.

adrianthedev:
I pushed a chunkier commit which updates Avo to `3.0.1.beta.7`. It's stable and ready for the public.
I'll push updates when we increment the version number to GA.

For full disclosure, the `searchable` field option is available in the `Pro` tier. I enabled it in the resource, and it will gracefully fallback to a regular select for the Community (free) tier which is installed by default.

I enabled the search for all resources.

The downside is that `ransack` now requires all models to have the `ransackable_attributes` method as a security measure.
I think you don't want to "muddle" up the models and offer clean models to those installing BT.

I have thought about what we could do here:

1. Use indirection and hide the methods under the `Base` modules
2. Use indirection and hide the methods under a different `Ransack`|`Search`|`Admin` concern
3. Add a new Concern and extend it from the `avo.rb` initializer

Example for 3.
```ruby
module UserExtensions
  extend ActiveSupport::Concern

  def self.ransackable_attributes()
    ...
  end
end
Rails.configuration.to_prepare do
  User.include UserExtensions
end
```

I would choose 1. but will follow your advice.

jagthedrummer:
Hey @adrianthedev, let me give this a little more thought. The added info about `searchable` only being available in Avo Pro has me reconsidering whether we should have that included and turned on by default. Let me think through the implications of that and I'll get back to you soon.

adrianthedev:
Generally, it takes a while until anyone needs the `searchable` option, and usually they have enough traction to switch over to Pro.

Sure, let me know if I can answer any more questions.

jagthedrummer:
Hey @adrianthedev, first of all, I'm sorry to have started all of the back and forth over this.

I'm just laying out my current thoughts here so maybe you and @andrewculver can help guide this PR in for a landing.

Given the new info that `searchable` is only available in the Pro package I'm now leaning towards the idea that we shouldn't directly include any of the searchable/ransack stuff. Just to keep everything as simple as possible for the functionality that we actually ship.

Whatever we decide on that question I think we'll want to add a page to the documentation site that lets people know that Avo is part of the BulletTrain package and how they can get into it. There we should also mention how they can configure their own custom models to be available to Avo (can/should we make super scaffolding do this automatically and/or with a dedicated task). As part of that page we could also include a mention of Avo Pro, tell people _why_ they might want to upgrade, and then have some instructions on _how_ to upgrade to get the most benefit. Like telling them they need to install `ransack` and configure it, and anything else that they might need to do. (It might me as simple as linking to some existing Avo Pro docs.)

I think then we might leave the choice of where to put the `ransackable_attributes` up to the developer. We could list the options that you described above, and give recommendations on how to decide which to use.

Again, this is just me kind of thinking out loud, and I'm open to suggestion/persuasion. What do y'all think?

adrianthedev:
Hey @jagthedrummer.

Apologies for the delay. I went on vacation and then the conference season started, and just about now I'm getting to the bottom of my backlog.

I couldn't agree more with what you wrote.
Let's just ship something that brings immediate value to users.
We have everything rather well-documented and they can find their way fast to the `searchable` docs page.

**Please direct me to the repository with the docs** and I'll gladly write a page about this integration and how users should continue with Avo and BT.

I'd love to span this integration further and automatically expose super scaffolding commands to Avo resources. That's the experience I'd love to give to users.

I removed the `ransack` bits and commented out the search functionality for now.
As far as I'm concerned this PR is just perfect for the initial iteration.

So yeah, looking forward to your feedback @jagthedrummer and @andrewculver.
Thank you ✌️

jagthedrummer:
@adrianthedev, now it's my turn to apologize for the long delay! 😬 Somehow this managed to sneak past me. I'll try to give this a good look later this week and hopefully we can get it merged.

To answer your question about the docs, they live in the `core` repo inside the main `bullet_train` gem: https://github.com/bullet-train-co/bullet_train-core/tree/main/bullet_train/docs

Those markdown files correspond directly to the pages on the docs site: https://bullettrain.co/docs

adrianthedev:
No worries @jagthedrummer. Thank you for taking the time to review it.

I'll make time in the next few days to update the docs too.

adrianthedev:
No worries @jagthedrummer. Thank you for taking the time to review it.

I'll make time in the next few days to update the docs too.

jagthedrummer:
@adrianthedev, would you mind doing a full rebase on this and making sure that you've updated `avo` in `Gemfile.lock` to the latest version?

As a last check before merging I pulled down this branch and rebased it, and rather than resolving conflicts in `Gemfile.lock` by hand I just checked out the HEAD version of that file and then did a `bundle install` which got me to `avo` 3.1.6. I pushed that to GitHub to let CI run, and CI is failing with a complaint about a `tabs_style` config option.

```
NoMethodError: undefined method `tabs_style=' for #<Avo::Configuration:...
# snip
>> (NoMethodError)

  config.tabs_style = :pills
        ^^^^^^^^^^^^^
/home/runner/work/bullet_train/bullet_train/tmp/starter/config/initializers/avo.rb:67
```

https://github.com/bullet-train-co/bullet_train/actions/runs/7198334742/job/19607520823?pr=1241

When I try to run the app locally it fails with the same error. I tried commenting out that line and then I get complaints about `license_key` and `fields`.

Seems that the initializer in this PR isn't up to date with the latest expectations of `avo`.

Relatedly, we have `depfu` set up on this repo to automatically bump gems if possible when new versions are released. Given that there's a decent amount of integration code in this PR should we pin `avo` to a more specific version?

I guess the real underlying question is more about the stability of the interfaces that are used in this PR. Do you expect them to be stable? In which case we could probably let the version float. Or are you still making developer-facing API level changes? In which case we should probably pin to at least a major version, if not minor.

adrianthedev:
Done! Rebased.

I'm not sure why the initializer was off like that (the `tabs_style` issue). Probably from a borked rebase.
It should be ok now.

Regarding breaking changes in the API, we do occasionally ship those. It's mostly to improve past mistakes.
With Avo 3 launched, we expect those to be far less than before as we made the API more consistent.
We have a thorough process of documenting breaking changes on our documentation website ([v2](https://docs.avohq.io/2.0/upgrade.html) and [v3](https://docs.avohq.io/3.0/upgrade.html)).
We also have a two-week release cycle where we document everything we ship.
We don't follow semver but increment the minor version every two weeks and the patch version only when we ship quick bug fixes and for security updates.

My recommendation is not to pin it down Avo because the configuration files that BT ships have a very established API that's not prone to changes.
We will monitor and bring updates as we document them, but as I said, I don't foresee changes to this API.

Avo being so "alive" with so many new features every month, pinning down the version will almost make unpinning the version a default for users who want to use it.

I will default to your choice, though.

config/initializers/avo.rb
```diff
@@ -0,0 +1,109 @@
+# For more information regarding these settings check out our docs https://docs.avohq.io
+Avo.configure do |config|
+  ## == Routing ==
+  config.root_path = "/avo"
```
jagthedrummer, line 4: I think it would be nice to move this into the `/developers/` path like we did with Sidekiq.

```suggestion
  config.root_path = "/developers/avo"
```

jagthedrummer (approved):
Looks good, @adrianthedev! I made one small suggestion about moving it into the `/developers/` "route space" that we're already using for the Sidekiq web UI. Does that seem reasonable to you? Otherwise, I think it's ready to go.

BTW, I think the problem with the `tabs_style` thing was my fault. I hadn't done a `git fetch` on your remote, and so when I checked out the branch I was getting something from some point in the past. Sorry about that!

adrianthedev:
Yes, I think it's reasonable. I clicked commit pretty fast, but now I'm even pondering the `/admin` route.
I think `/developers` route might give some non-developer folks the heebie-jeebies that maybe they are not allowed there or that maybe they should mess things up.

I mean, my vision for this is to really enable the whole team (not just devs) to use it in production.

What do you think?

jagthedrummer:
Well, on the one hand someone absolutely _could_ mess things up via `avo`...

But I do see what you mean about it being aimed at a larger group than strictly "developers".

🤔

config/initializers/avo.rb
```diff
@@ -0,0 +1,109 @@
+# For more information regarding these settings check out our docs https://docs.avohq.io
+Avo.configure do |config|
+  ## == Routing ==
+  config.root_path = "/developers/avo"
```
jagthedrummer, line 4: OK, @adrianthedev, let's go with `/admin` as the prefix. I think that's a good suggestion since it's not exclusively a developer concern in the same way that Sidekiq (probably) is.

```suggestion
  config.root_path = "/admin/avo"
```

jagthedrummer:
Thanks for the PR, @adrianthedev! I agree that we should not pin `avo` in the `Gemfile`. Gonna merge it and release a new version. 🚀

adrianthedev:
@jagthedrummer thank you for your time and support with this PR!
It means a lot for us to have Avo as the default admin panel in Bullet Train.

One note on the path. What do you think about having just `/admin` point to Avo directly.
I think people are used to (and some expect it) to have the `/admin` path take them to the admin panel.

I'm going to get started on those docs.

Thanks again @jagthedrummer and @andrewculver for your support!

jagthedrummer:
@adrianthedev, I had considered just `/admin` but then thought "what if there are other admin type things that we/developers would like to put into the admin route space?" Maybe we should put a landing page at `/admin` that has a link to `/admin/avo`. Or maybe just a redirect for now.

adrianthedev:
That's a good question! Not really so obvious.
There are a couple of scenarios:

1. You are using the Community (free) tier

You can pass the project without any limitations to the customer as Avo Community is [LGPL v3](https://opensource.org/license/lgpl-3-0), an Open Source Initiative approved license

2. You are using Avo on the Pro or Advanced tiers

You can pass that license to your customer where they can take ownership of it and maintain the subscription in order to get updates.

If they don't want to maintain the subscription, they get to keep the version that they are on indefinitely.

So you, and your customer should be future-proofed in terms of licensing.


I'm here if I can answer any other questions.

PR #797, opened 2023-05-19
Check dependencies for Linux in bin/configure

gazayas:
This PR ensures we have the proper dependencies installed when running `bin/configure` on Linux. It's not as simple as running `check_package` like we do with MacOS, but grabbing the system packages with `dpkg` lets us know if the package is installed or not as long as we know the proper name for it.

I didn't bother too much with `node` because #786 is open, but we can figure out the details if we get a merge conflict later on.

### DIFF
--- a/bin/configure
+++ b/bin/configure
@@ -142,14 +142,6 @@ else
   end
 end
 
-if `yarn -v 2> /dev/null`.length > 0
-  puts "Yarn is installed.".green
-else
-  puts "You don't have Yarn installed. We can't proceed without it. Try `brew install yarn` or see the installation instructions at https://yarnpkg.com/getting-started/install .".red
-  exit
-end
-
-
 required_node = `cat ./.nvmrc`.strip
 actual_node = begin
                 `node -v`.strip.gsub("v", "")


jagthedrummer:
@gazayas I just merged #786 and, as you predicted, there are merge conflicts on this one now.

gazayas:
@jagthedrummer Done!

This is what it looks like on Linux by the way. Looks like I need to update PostgreSQL :sweat_smile:

Screenshot from 2023-08-03 15-37-38

bin/configure
```diff
@@ -87,9 +87,68 @@ else
...
+    exit
+  end
+end
+
+if `yarn -v 2> /dev/null`.length > 0
+  puts "Yarn is installed.".green
+else
+  puts "You don't have Yarn installed. We can't proceed without it. Try `brew install yarn` or see the installation instructions at https://yarnpkg.com/getting-started/install .".red
+  exit
+end
```
jagthedrummer, line 150: @gazayas This whole block from lines 145-150 is a duplicate of code that already exists lower down in this file. (Lines 173-178 in your new version of the file, and/or lines 114-119 in the original file.) I think we can remove it from this PR.

jagthedrummer (changes requested):
One little bit of duplication seems to have snuck in during merge conflict resolution.

bin/configure
gazayas: @jagthedrummer Done!

jagthedrummer (approved):
Looking good!

PR #799, opened 2023-05-29
Add Super Select helper to customize options

### LINKED ISSUE #783: Declare Super Select options in the starter repository
I recently thought back on this conversation on Discord where a developer was trying to edit their super select partial: https://discord.com/channels/836637622432170028/1083307718808453150

Basically they wanted to have a blank field so they used `options: {include_blank: true}`, but it ended up adding some unnecessary space:
image

After searching around a bit for a good answer, I came across this [Stack Overflow](https://stackoverflow.com/questions/22410180/how-to-allow-user-to-select-empty-string-on-select2) post which reminded me that we declare options for select2 elements in our [super select stimulus controller](https://github.com/bullet-train-co/bullet_train-core/blob/20ec101e9bb23f0858c00b2976edd6633bb8421a/bullet_train-fields/app/javascript/controllers/fields/super_select_controller.js#L46-L48), not one of the partials:
```ruby
let options = {
  dropdownParent: $(this.element)
};
```

I propose adding either a new file or some logic in `application.js` which allows users to declare their options for Super Select elements which are then read and implemented in our `bullet_train-core`-side stimulus controller.

gazayas:
Closes #783.

Details in Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/310

### DIFF
--- a/app/helpers/super_select_helper.rb
+++ b/app/helpers/super_select_helper.rb
@@ -1,0 +1,10 @@
+module SuperSelectHelper
+  # Place your options for customing super select elements here.
+  # https://select2.org/configuration/options-api
+  def super_select_config
+    {
+      # allowClear: true,
+      # dropdownCssClass: "color: blue;"
+    }.to_json
+  end
+end


gazayas:
Closing this since I decided to pass options via the render call which is applied in the partial in https://github.com/bullet-train-co/bullet_train-core/pull/310

PR #800, opened 2023-05-29
Remove version restriction for the debug gem

gazayas:
The issue over at `debug` is now closed and we're using Ruby 3.2.2, so I removed the version restriction.


jagthedrummer (approved):
👍

PR #801, opened 2023-06-08
API Mode

### LINKED ISSUE #790: Proposal to introduce API only mode
We can do the following to create an api-only application in rails:
```
> rails new application_name --api
```

Since we have a lot of powerful API logic in Bullet Train, I thought it would make sense for developers to run one of the following two commands to setup their application for API mode.

## 1. `bin/configure --api`
After cloning the repository, developers would run `bin/configure --api` to remove any unnecessary files so they only have what they need for an backend API. This feels semi-unintuitive because rails creates an api-only application on `rails new`. However, I think using our equivalent to `rails new` presents some of its own issues, which I'll mention here next.

## 2. `git clone`

```
> git clone --branch api-only git@github.com:bullet-train-co/bullet_train.git application_name
```

Here we would maintain an `api-only` branch which would look similar to `main`, except that it would only have the files necessary for an api-only application. The problem is that I feel like it would be pretty hard to maintain one apart from `main`, and it would also be pretty tough to just make an altogether new, different `bullet_train` API starter repository just for this purpose. Yes, it's similar to `rails new` which would make sense for most developers upfront, but if all we have to do is add `--api` for `bin/configure`, then that's probably a better option when considering all of the files we have to maintain.

## Afterthoughts
If we <i>were</i> to implement this, I think `bin/configure --api` would be all-around easier to maintain. However, if we already have a way to implement an api-only application that I'm just missing, I'd be glad to re-think through things.

gazayas:
Closes #790.

Getting a head start on implementing API mode, and I just started off by removing files I don't believe we'll need.

I think we'll need to remove specific gems and run `bundle install` as well.

Tagging you @newstler since I saw you assigned yourself to #790. I also don't entirely have the end-goal envisioned yet either, so I think it'll take some trial and error to figure out what we really need or don't need.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -134,6 +134,9 @@ group :development do
 
   # For colorizing text in command line scripts.
   gem "colorize"
+
+  # Allow users to document and showcase their partials, components, view helpers, etc.
+  gem "showcase-rails"
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -649,6 +649,7 @@ DEPENDENCIES
   sentry-rails
   sentry-ruby
   sentry-sidekiq
+  showcase-rails
   simplecov
   sprockets-rails
   standard
--- a/bin/configure
+++ b/bin/configure
@@ -332,6 +332,15 @@ def replace_content
 
   CODE
   replace_in_file("./config/environments/production.rb", code_to_remove, "")
+
+  code_to_remove = <<-CODE
+  system("yarn check") || system!("yarn install")
+  system!("bin/link")
+
+  puts "== Building assets =="
+  system!("yarn run build && yarn run build:css && yarn light:build:css")
+  CODE
+  replace_in_file("./bin/setup", code_to_remove, "")
 end
 
 def edit_ci!
--- a/bin/configure
+++ b/bin/configure
@@ -265,6 +265,7 @@ def replace_content
   puts ""
   puts "Replacing assets path with api-only configuration in application.rb.".green
   assets_path = 'config.assets.paths << Rails.root.join("app", "assets", "fonts")'
+
   api_only = <<-CODE.chop
 
     # Only loads a smaller set of middleware suitable for API only apps.
@@ -272,7 +273,10 @@ def replace_content
     # Skip views, helpers and assets when generating a new resource.
     config.api_only = true
   CODE
-  replace_in_file("./config/application.rb", assets_path, api_only)
+  # TODO: I would like to implement this, but it's causing some tests to fail
+  # saying we need to provide a session to use OmniAuth.
+  # replace_in_file("./config/application.rb", assets_path, api_only)
+  replace_in_file("./config/application.rb", assets_path, "")
 
   # config/environments/development.rb
   puts ""
--- a/bin/configure
+++ b/bin/configure
@@ -244,7 +244,7 @@ end
 if options[:api]
   require 'fileutils'
 
-  puts "Removing the following files.".blue
+  puts "Removing files unneeded in API mode.".blue
   removal_list = []
   removal_list << Dir.glob("app/controllers/**").reject {|dir| dir.match?(/[api|rb]$/)}
   removal_list << "app/javascript"
@@ -257,6 +257,42 @@ if options[:api]
   removal_list.each do |file_or_dir|
     puts FileUtils.rm_rf(file_or_dir)
   end
+
+  puts "Editing the CircleCI test flow for API mode.".blue
+  new_lines = []
+  File.open(".circleci/config.yml", "r") do |file|
+    lines = file.readlines
+
+    # Skip these lines inbetween these indices.
+    start = nil
+    finish = nil
+
+    # Also skip unneeded test invocations after workflow declaration.
+    workflow_start = nil
+
+    lines.each_with_index do |line, idx|
+      if line.match?(/\s*'Minitest for Super Scaffolding'/)
+        start ||= idx
+      elsif line.match?("test/system/action_models_test.rb")
+        finish ||= idx
+      elsif line.match?(/^workflows/)
+        workflow_start = idx
+      end
+    end
+    tests_to_skip = (start..finish)
+
+    new_lines = lines.map.with_index do |line, idx|
+      if tests_to_skip.include?(idx)
+        next
+      elsif idx >= workflow_start && (line.match?(/System/) || line.match?(/Scaffolding/))
+        next
+      else
+        line
+      end
+    end.compact
+  end
+
+  File.write(".circleci/config.yml", new_lines.join)
 end
 
 puts ""
--- a/bin/configure
+++ b/bin/configure
@@ -1,25 +1,404 @@
 #!/usr/bin/env ruby
 
-puts ""
-puts `gem install colorize`
-puts ""
-
-puts ""
-puts `gem install activesupport`
-puts ""
-
 require 'optparse'
 require 'colorize'
 require 'active_support'
+require 'yaml'
+require 'fileutils'
 
-options = {}
-OptionParser.new do |opts|
-  opts.banner = "Usage: bin/configure [options]"
+##############################
+# Main `step` methods
+##############################
 
-  opts.on("-a", "--api", "Create application in API mode") do |api|
-    options[:api] = api
+def install_necessary_gems
+  puts ""
+  puts `gem install colorize -N`
+  puts ""
+
+  puts ""
+  puts `gem install activesupport -N`
+  puts ""
+end
+
+def check_options
+  @options = {}
+  OptionParser.new do |opts|
+    opts.banner = "Usage: bin/configure [options]"
+
+    opts.on("-a", "--api", "Create application in API mode") do
+      @options[:mode] = "api"
+    end
+    opts.on("-u", "--ui", "Create application in UI mode") do
+      @options[:mode] = "ui"
+    end
+
+  end.parse!
+
+  @modes_config = YAML.load_file('./config/modes.yml') if @options[:mode] # Load the yaml config
+end
+
+def check_ruby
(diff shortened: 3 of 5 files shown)


newstler:
Hey, I had a bit different approach: I placed yml file in `config` folder with the lists of files and gems needed for API and UI only modes, and deleted stuff according to those files, and after deleted this yml config. Not sure which approach is better? Love that you did also CI modification!

gazayas:
Having a yml file to keep track of everything works for me too! We're you thinking of loading the files and directories through that, and then using `FileUtils.rm_rf` to remove them like I have here?

Feel free to push any changes directly onto this branch, I'd be glad to work with you on it.

newstler:
Seems I can't push into your repo so I had to create [this pull request](https://github.com/gazayas/bullet_train/pull/1), yet it is still work in progress.

gazayas:
I added a few more commits.

1. `capybara-email` was still being included which is what was giving us the error telling us to run `bundle update capybara`, so I added it to `modes.yml`.
2. I removed the `bullet_train` gem from `modes.yml` so we could include it in the Gemfile. Without the gem, I found that we were getting a lot of errors due to helper methods in `bullet_train-core/bullet_train/lib/bullet_train.rb` that we need.
3. I removed theme-related initializers that were also throwing errors since `BulletTrain::Themes` wasn't defined.

We're still getting an error because `showcase-rails` is a dependency in `bullet_train-core/bullet_train`. We might need to make a PR over there to add `showcase-rails`, but just not make it a dependency.

## `bin/setup`
We need to do some work in `bin/setup` to because it has some lines that call yarn and `bin/link`:

https://github.com/bullet-train-co/bullet_train/blob/56914548b0a2718300eaa3a733eb9b168afb9214/bin/setup#L35-L39

gazayas:
Besides that, there are some tests that aren't passing now, but I think this takes care of most of what we want to do. I'll leave it for a bit and see if we have to edit the tests themselves or edit `modes.yml` to include the right gems/files.

gazayas:
Ok, I got some more tests to work.

# OmniAuth needs a session to work
I know since we want to use a REST API there should be no reason to use a session, but for some reason with the way our API controller tests are set up, OmniAuth needs a session for these tests to work.

You can see this failure if you put `config.api_only = true` in `config/application.rb` and run the API teams controller test:

```
Error:
Api::V1::TeamsControllerTest#test_update:
OmniAuth::NoSessionError: You must provide a session to use OmniAuth.
    test/controllers/api/test.rb:10:in `access_token'
    test/controllers/api/v1/teams_controller_test.rb:59:in `block in <class:TeamsControllerTest>'
```

# Other necessary files
There were some other files that I removed from `modes.yml` because I believe we need them. For example, if we're using OmniAuth, we might need `app/controllers/account/oauth/`. We were an error due to the fact that the application helper had been deleted.

# Questions from me
Besides that, I think I want to wait for a reply concerning the following two files:
1. `test/controllers/api/open_api_controller_test.rb`
2. `test/models/invitation_test.rb`

The Open API controller test because it has this line, but we remove `yarn` so this test fails:

https://github.com/bullet-train-co/bullet_train/blob/56914548b0a2718300eaa3a733eb9b168afb9214/test/controllers/api/open_api_controller_test.rb#L11

Also I'd like to know what to do about the `Invitation` model test because we get a missing template error concerning the mailer:
```
Error:
InvitationTest#test_must_be_valid:
RuntimeError: Neutered Exception ActionView::Template::Error: Missing partial themes/light/layouts/_mailer with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]}.
```

# Conclusion
And that's it from me, I think it'll be difficult to do anything else unless we get more direction.

gazayas:
With the following PR, I removed showcase-rails from bullet_train and added it here directly:
- https://github.com/bullet-train-co/bullet_train-core/pull/322

gazayas:
By the way, I haven't committed this, but I rewrote `remove_gems!` with [Masamune](https://github.com/gazayas/masamune-ast) and it looks like this:

```ruby
def remove_gems!
  gems_to_remove = @modes_config[@options[:mode]]['gems'] # Extract the gem names
  gemfile_lines = File.readlines('./Gemfile') # Read the Gemfile

  msmn = Masamune::AbstractSyntaxTree.new(gemfile_lines.join)
  gems_to_remove = msmn.strings.select{ |str| gems_to_remove.include?(str[:token]) }
  gem_line_numbers = gems_to_remove.map { |gem| gem[:position].first }

  # Comments sit one line above their gem.
  # If the next line is a gem we want to remove,
  # we get the comment line to remove as well.
  comments = msmn.comments
  comments_to_remove = comments.select do |comment|
    comment_line_number = comment[:position].first
    gem_line_numbers.include?(comment_line_number + 1)
  end
  comment_line_numbers = comments_to_remove.map { |comment| comment[:position].first }

  lines_to_remove = gem_line_numbers + comment_line_numbers
  filtered_gemfile_lines = gemfile_lines.reject.with_index { |_, index| lines_to_remove.include?(index) }

  puts "Removing gems unneeded in #{@options[:mode].upcase} mode.".blue

  # Write the new Gemfile
  File.open('./Gemfile', 'w') do |file|
    file.puts filtered_gemfile_lines
  end
end
```

newstler:
Hi, sorry for being out, was moving to a new apartment and there's no internet here still :)

Great work!
I have several thoughts though if I may:

### bullet_train gem
I will check which helper methods in `bullet_train-core/bullet_train/lib/bullet_train.rb` were showing exceptions, but it feels like they should be extracted to a separate gem under `bullet_train-core`.

### showcase-rails
It's a totally UI-oriented gem, so definitely we don't need it in API mode.

### bin/setup
As this script runs _before_ `bin/configure`, shouldn't we move all the UI/API modes logic to it? That would let us remove this line with `yarn run...`. Also, it kinda feels more natural to do all our manipulations there.
Btw, I am aware why those two scripts — setup and configure — are separated? Because there are no options (till now) to each of them and they are always being run one after another?

### OmniAuth
To access API we use Doorkeeper with Bearer token (issued for `Platform::Application`). As I know, it is the only way to use API. So, why do we need OmniAuth then?
I will take a closer look on what happens when `post "/oauth/token"`, why this route is used in test (feels like should be something like `api/teams/{team_id}/platform/applications/access_tokens`), and why it calls OmniAuth.

### test/controllers/api/open_api_controller_test.rb
To be honest, not sure if it's a good idea to install yarn just for this test, but this one actually tests API documentation :)
But I remember looking to replace redocly due to some "bugs" or better to say differences how it and bullet train understands routes (see [here)](https://github.com/bullet-train-co/bullet_train/issues/734#issuecomment-1528850290).

### test/models/invitation_test.rb
Not sure why mailer partials are missing, have they being deleted? Will take a look.

### remove_gems! with Masamune
Any reason to use source code analyzer here? For a common developer it kinda makes the code less readable and increases the learning curve, whilst it doesn't seem to simplify it or make it shorter...
There were also a couple of comments that had to stay:
```
        !gemfile_lines[i].include?("Core packages.") &&
        !gemfile_lines[i].include?("Optional support packages.")
```

### Conclusion
More and more I think that this could be done better. I see it as:
```
> gem install bullet_train
...
> bullet_train new my_app --api
... # here all the setup and configuration happens
```
I researched and it doesn't feel like impossible. Any objections?

gazayas:
@newstler

# bin/setup
`bin/setup` should actually be run [after bin/configure](https://github.com/bullet-train-co/bullet_train/issues/573#issuecomment-1373510557), and this came up recently in [Discord](https://discord.com/channels/836637622432170028/1116294694599663658/1116322544253947944), so I'm ok with leaving all of this logic in `bin/configure`.

> Btw, I am aware why those two scripts — setup and configure — are separated?

Sorry, I'm not sure!

# OmniAuth in API tests

> To access API we use Doorkeeper with Bearer token (issued for Platform::Application). As I know, it is the only way to use API. So, why do we need OmniAuth then?
I will take a closer look on what happens when post "/oauth/token", why this route is used in test (feels like should be something like /api/teams/{team_id}/platform/applications/access_tokens), and why it calls OmniAuth.

Yeah I'm not entirely sure myself and I haven't taken the time to look into it, but like you said, we use a Bearer token so I think it makes the most sense to edit these tests.

# Masamune

> Any reason to use source code analyzer here? For my point of view it makes the code less readable and increases the learning curve, whilst it doesn't seem to simplify it or make it shorter...

As a whole I want to start using Masamune for code replacement logic since I think it will be more precise in grabbing parts of code we want to edit, but I didn't actually commit the code because its more of a proposal than anything. I haven't talked to Andrew about it but I don't plan on actually committing anything here until we decide if we want to use the gem or not.

newstler:
### test/models/invitation_test.rb
It appears it fails because it uses verb view from `bullet_train-themes-light` gem, which uses such assets as `images/logo.png` and `css/application.css`, as well as some of dependencies needs `assets/config/manifest.js`.

Then I have a question: invitations send emails with a link to UI where a new user may register. But we don't have UI anymore, so this link would be broken anyway. So for API mode I see two options: either delete Invitations functionality completely, or modify it in API mode so emails send API key. Which is quite a different way of how invitations should work, and maybe that would better to create some parallel code like `ApiInvitation`...

### OmniAuth
As an API-only app, we're using Doorkeeper to authenticate "applications", each client application has its own credentials (client ID and client secret), and it uses these credentials to obtain an access token from our Doorkeeper-powered API. This access token is then used to authenticate subsequent requests to the API. In this flow, individual users never interact with API directly, so there's no need for a User model.

So the solution here would be to delete User model, devise gem and everything that uses omniauth (bullet_train-integrations-stripe for example). The hardest would be deleting User model I assume as a lot in the codebase of starter app and bullet train gems should depend on it.

### Separate repo?
Here I come back to the thought about separate repo for starter app and bullet train gems affected. What do you think, won't it make more sense now? @gazayas

PR #811, opened 2023-07-15
Add tests for localized date and datetime fields

3pns:


jagthedrummer:
@3pns is this dependent on getting the themes localization stuff merged? (If so, would it be worth adding to the branch that you're using for cross repo testing of that stuff?)

3pns:
@jagthedrummer Yes I am currently working on it in #845. I will close this branch afterward

PR #696, opened 2023-03-14
Update dates helper test

gazayas:
Here are the tests for this pull request
- https://github.com/bullet-train-co/bullet_train-core/pull/182

I deleted some of the assertions because we deleted the following over in there:
```ruby
if local_time(timestamp).to_date == local_time(Time.now).to_date
  "Today at #{display_time(timestamp, custom_time_format)}"
# yesterday?
elsif (local_time(timestamp).to_date) == (local_time(Time.now).to_date - 1.day)
  "Yesterday at #{display_time(timestamp, custom_time_format)}"
else
  # ...
end
```

PR #828, opened 2023-08-08
Update dates helper

gazayas:
Continuation of #696.

Submitting as draft since I still have a little bit of work to do in the joint PR, https://github.com/bullet-train-co/bullet_train-core/pull/182

### DIFF
--- a/test/system/dates_helper_test.rb
+++ b/test/system/dates_helper_test.rb
@@ -43,20 +43,12 @@ class DatesHelperTest < ApplicationSystemTestCase
         click_on "Back"
 
         # Assert today's date is displayed correctly.
-        assert page.has_text? "Today at #{time.strftime("%l:%M %p").strip}"
+        assert_text I18n.l(time, format: :default)
 
         # Assert yesterday's date is displayed correctly.
         travel_to time + 1.day
         visit current_url # Refresh the page
-        assert page.has_text? "Yesterday at #{time.strftime("%l:%M %p").strip}"
-
-        # Assert the month and day is shown for anything before then.
-        travel_to time + 2.days
-        visit current_url
-
-        # We have to assert these two things separately so it doesn't fail on the last day of the year when the year is present.
-        assert page.has_text? time.strftime("%B %-d").strip.to_s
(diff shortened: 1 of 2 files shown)


gazayas:
@jagthedrummer I updated the tests to use `page.execute_script` from #811 for the new date-related fields.

Besides that, I got the Super Scaffolding tests to pass by using a fix to transformer which I temporarily added to https://github.com/bullet-train-co/bullet_train-core/pull/182. I'll revert that here now that the tests are passing.

The transformer fix can be found here: https://github.com/bullet-train-co/bullet_train-core/pull/386

gazayas:
@kaspth I also went ahead and used `assert_text` like you mentioned [here](https://github.com/bullet-train-co/bullet_train/pull/696#discussion_r1286082258) 👍

jagthedrummer:
@gazayas After merging this we have one failure in CI when running against the released version of the BT gems. And I didn't realize until after I'd merged it that this branch hadn't run against both the released version and `core/main`.

The failure is:
```
expected to find text "Thu, 10 Aug 2023 14:52:00 +0000"
```

And what's actually there is:
```
Today at 2:52
```

Is that expected? And is it what we want? `Today at 2:52` seems much more user friendly to me than `Thu, 10 Aug 2023 14:52:00 +0000`.

gazayas:
@jagthedrummer About the strings like `"Today at 2:52"`, I expressed [here](https://github.com/bullet-train-co/bullet_train-core/pull/182#issue-1620000994) that we would trade this off in favor of the Rails default:

> We unfortunately lose the "Today at ..." and "Yesterday at ..." strings we once had, but I think this is a good trade off because we can then write our partials like this:
```erb
<%= render 'shared/attributes/date', attribute: :test_date %> <!-- Uses default format -->
OR
<%= render 'shared/attributes/date', attribute: :test_date, format: :short %>
OR
<%= render 'shared/attributes/date', attribute: :test_date, format: '%m/%d/%Y' %>
```

Because of that, I was under the assumption that we would be ok with the change. What do you think is the best approach? Would be glad to look over the tests or whatever's necessary.

gazayas:
@jagthedrummer, I saw that you opened #844, LGTM

PR #831, opened 2023-08-08
Run tests against released gems and `main` in `core`

jagthedrummer:
This makes it so that every PR and every merge to main will run tests in CI against both the released versions of the `core` gems and against a branch in the `core` repo.

For merges to `main` in the starter repo, they'll always run against the latest state of `main` in `core`.

For PRs we'll attempt to find a branch in `core` that matches the name of the branch being tested. If we can find one, we'll test against that. If not we'll test the PR branch against `main` in `core`.

The tests that we run against the released versions of the gems can probably be expected to fail with some regularity since we sometimes introduce changes into the starter repo that depend on particular changes in `core`. But this will help us to know when we've made changes in the starter repo that mean that we need to relase a new version of the `core` gems.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,10 @@ jobs:
       - restore_cache: *restore_bundler_cache
       - restore_cache: *restore_yarn_cache
 
-      - run: ./bin/checkout-and-link-core-repo
+      - when:
+          condition: << parameters.use-core-repo >>
+          steps:
+            - run: ./bin/checkout-and-link-core-repo
 
       # Install dependencies
       - ruby/bundle-install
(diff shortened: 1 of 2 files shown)


gazayas:
> After getting this working we can see that there are currently failures in the Super Scaffolding tests when we run them against the main branch of core, but not when we run them against the released gems (which was the only thing we used to do previously).

Yes, I wrote a comment not to long ago [here](https://github.com/bullet-train-co/bullet_train-core/pull/386#issuecomment-1670476833), linking it so we know it's related.

jagthedrummer:
I'm going to go ahead and merge this. If there are objections we can always revert.

kaspth:
Seems good to me! I don't know if these updates mean that Bullet Train apps will also try running the `core` branches CI jobs after they pull this commit?

jagthedrummer:
> I don't know if these updates mean that Bullet Train apps will also try running the `core` branches CI jobs after they pull this commit?

That's... a very good point that I didn't think about. I was thinking of the CI config as only being for the starter repo itself, but yeah. I guess if other devs were relying on that config then they'd get the extra tests as well. 🤔

kaspth:
Yeah, not sure how we handle that, but from the test failures you've uncovered we definitely need this, so I say ok for now!

PR #834, opened 2023-08-10
Move `collection_actions` into `RoutesHelper`

gazayas:
Continuation of #21.

### DIFF
--- a/app/helpers/routes_helper.rb
+++ b/app/helpers/routes_helper.rb
@@ -1,8 +1,0 @@
-module RoutesHelper
-  # `collection_actions` is automatically super scaffolded to your routes file when creating certain objects.
-  # This is helpful to have around when working with shallow routes and complicated model namespacing. We don't use this
-  # by default, but sometimes Super Scaffolding will generate routes that use this for `only` and `except` options.
-  def self.collection_actions
-    [:index, :new, :create]
-  end
-end
--- a/app/helpers/routes_helper.rb
+++ b/app/helpers/routes_helper.rb
@@ -1,0 +1,8 @@
+module RoutesHelper
+  # `collection_actions` is automatically super scaffolded to your routes file when creating certain objects.
+  # This is helpful to have around when working with shallow routes and complicated model namespacing. We don't use this
+  # by default, but sometimes Super Scaffolding will generate routes that use this for `only` and `except` options.
+  def self.collection_actions
+    [:index, :new, :create]
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,11 +4,6 @@ Rails.application.routes.draw do
   draw "devise"
   draw "sidekiq"
 
-  # This is helpful to have around when working with shallow routes and complicated model namespacing. We don't use this
-  # by default, but sometimes Super Scaffolding will generate routes that use this for `only` and `except` options.
-  # TODO Would love to get this out of the application routes file.
-  collection_actions = [:index, :new, :create] # standard:disable Lint/UselessAssignment
-
   # This helps mark `resources` definitions below as not actually defining the routes for a given resource, but just
   # making it possible for developers to extend definitions that are already defined by the `bullet_train` Ruby gem.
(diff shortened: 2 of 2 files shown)


gazayas:
The tests are failing on the super scaffolding system test setup script, not on the actual test itself. Not sure what's going on there. The other nodes for the same test are passing though.

jagthedrummer:
@gazayas The failure is specifically complaining that it's missing the variable that you moved. It's saying `undefined local variable or method 'collection_actions' for #<ActionDispatch::Routing::Mapper:0x00007f307371f840...`.


```
/home/circleci/project/config/routes.rb:67:in `block (6 levels) in <main>':
undefined local variable or method `collection_actions' for
#<ActionDispatch::Routing::Mapper:0x00007f307371f840 @set=#<ActionDispatch::Routing::RouteSet:0x00007f3073abaf18>, @draw_paths=[#<Pathname:/home/circleci/project/config/routes>], @scope=#<ActionDispatch::Routing::Mapper::Scope:0x00007f3073a02cd8 @hash={:path_names=>{:new=>"new", :edit=>"edit"}}, @parent=#<ActionDispatch::Routing::Mapper::Scope:0x00007f308490f4a0 @hash=nil, @parent=nil, @scope_level=nil>, @scope_level=nil>, @concerns={:sortable=>#<Proc:0x00007f3073a02008 /home/circleci/project/vendor/bundle/ruby/3.2.0/gems/actionpack-7.0.7/lib/action_dispatch/routing/mapper.rb:2072 (lambda)>}, @devise_finalized=false> (NameError)
        from /home/circleci/project/vendor/bundle/ruby/3.2.0/gems/actionpack-7.0.7/lib/action_dispatch/routing/mapper.rb:871:in `scope'
        from /home/circleci/project/config/routes.rb:66:in `block (5 levels) in <main>'
```

I think that `RoutesHelper` has to be somehow configured to be available in `routes.rb` but I forget the incantation for that. Or maybe it has to be `require`d?

jagthedrummer:
@gazayas after thinking about this a bit, I'm starting to think that the best thing to do here might be to leave `collection_actions` in `routes.rb` and just remove the TODO.

In my mind `routes.rb` is something that I generally expect to be "self contained". I think that if we're going to generate code into `routes.rb` that references a variable called `collection_actions` that it would be helpful to devs if that variable were declared directly in `routes.rb`. Otherwise when a dev comes across a reference to that variable they have a "what the heck is this?" moment, and have to start grepping the codebase to find it.

I'm not 100% convinced I'm right on that, so I'm open to other opinions, and I won't prevent merging if you want to get the tests passing. Just sharing my latest thoughts.

gazayas:
@jagthedrummer
> Otherwise when a dev comes across a reference to that variable they have a "what the heck is this?" moment, and have to start grepping the codebase to find it.

Yup, makes sense to me.

The original comment states the following:
> This is helpful to have around when working with shallow routes and complicated model namespacing.

I'm honestly not entirely sure if we need `collection_actions` to be explicitly stated in `routes.rb`, and we probably don't have to tell the dev all that we're doing under the hood either. It would probably just be best to super scaffold `:index`, `:new` etc. as symbol literals and just make sure we're generating the correct symbols according to whatever namespacing we're working with in the `RoutesFileManipulator` or transformer. Again, I don't think it's something we need to explicitly comment on on the starter repository side.

<i>However</i>, I do believe that that would be the scope of another PR, so I'm fine with just updating the comment here to make things more clearer for devs reading the source code. What do you think?

gazayas:
So yeah, I just added a comment to make things clearer.

jagthedrummer:
> It would probably just be best to super scaffold `:index`, `:new` etc. as symbol literals and just make sure we're generating the correct symbols according to whatever namespacing we're working with in the `RoutesFileManipulator` or transformer.

@gazayas I agree with you 100% on this. And also agree that it's a different issue/PR. And I think the comment you added is better than the TODO just hanging out there.

gazayas:
🙇‍♂️

PR #835, opened 2023-08-10
[Pre-release] Enable bulk invitations on onboarding step

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/394

**Note:**
This PR is shipping a new feature that will be disabled by default. The feature is only partially complete, but we want to get this merged so that other team members can work on additional parts of the feature.

If you'd like to opt-in to this work-in-progress feature, you can add a line to `config/initializers/bullet_train.rb` like this:

```ruby
  # Enable bulk invitations on the user onboarding step.
  config.enable_bulk_invitations = true
```


gazayas:
All of the core tests are working! This one is ready for review now.

test/system/invitations_test.rb
```diff
@@ -178,13 +179,43 @@ class InvitationDetailsTest < ApplicationSystemTestCase
       fill_in "Last Name", with: "Yamaguchi"
       click_on "Next"
 
+      # Click on next to show that bulk invitations will raise an error if not filled out properly.
```
gazayas, line 182: We can potentially put all of the testing for bulk invitations in a new file like `invitation_details_test.rb` has separate tests, but I'm okay with this for now.

gazayas: Okay, I just went ahead and put the tests in their own file, `invitation_lists_test.rb`

jagthedrummer:
@gazayas this looks like a super handy feature, but I wonder if we should make a config setting that controls whether it's active or not. It's a pretty big new change and I'm not sure we want to force it on people who have an existing app running. If they have custom tests that cover sign-up & any custom onboarding flows that they've built then this change would break all of those tests.

Maybe we could add a toggle to `config/initializers/bullet_train.rb` and have it default to being turned off? (And possibly have `bin/configure` toggle it to on so that new apps get it by default?)

gazayas:
@jagthedrummer Done!

jagthedrummer:
@gazayas, that looks great! There's just one test failing both on the `Core:` tests here, and on the joint PR in `core`. Once that's fixed up I think this should be good to go.

gazayas:
@jagthedrummer Hmm, I'm not quite sure what's going on with the OpenAPI test, but that's the only one that's failing when HIDE_THINGS is true. It looks like we have `null: true` in the migration?

https://github.com/bullet-train-co/bullet_train/blob/1bc50c3aa4adf492d36bcd1d93aa174895fb3a8b/db/migrate/20200219013834_add_added_by_to_membership.rb#L3

I'm also not entirely sure what removing TangibleThings has to do with this 🤔 Any thoughts?

I got the other tests to pass, as well as the tests in the joint PR by the way.

gazayas:
I'll work on the merge conflict here, I also saw that https://github.com/bullet-train-co/bullet_train/issues/912 was opened though so I might wait until we resolve that because the only test that's failing here is the OpenAPI test for Core: Minitest with HIDE_THINGS.

gazayas:
Yeah, since the OpenAPI stuff is revolving around Invitations API and the subject of this PR is invitations, I'll go ahead and wait a bit.

gazayas:
@jagthedrummer Besides the release tests (as expected), the tests are passing!

test/system/invitation_lists_test.rb
```diff
@@ -0,0 +1,64 @@
+require "application_system_test_case"
+
+class InvitationListsTest < ApplicationSystemTestCase
+  if bulk_invitations_enabled?
```
jagthedrummer, line 4: This test isn't running because we default to this feature being turned off. I think this test should change the config setting prior to test setup to ensure that this test _always_ run. (And it should restore the original setting afterwards.)

jagthedrummer (changes requested):
@gazayas I just pulled this branch down and gave it a run. I have a few more suggestions that I'm going to put here, just so its' all in one place. I realize some of the actual changes may need to happen in the joint PR in `core`.

1. Let's change `Role ids` to just `Role` on the form.
CleanShot 2023-08-28 at 11 21 39

2. I think it should be possible to bulk-invite fewer than three people on this screen. What if there are only one or two other people on my team?

Here I try to submit a single invitation:
CleanShot 2023-08-28 at 11 24 40

And it complains about all three being empty, and it doesn't retain the one email address and role that I did submit.
CleanShot 2023-08-28 at 11 24 53

In the console I see three of this error:

```
Unpermitted parameter: :role_ids.
```

3. As I noted in an inline comment, the test for not skipping this screen isn't being run in CI because the feature defaults to being turned off. I think we can manually set that config in the test setup, and then restore it during teardown.

gazayas:
@jagthedrummer Thanks for the review! Here are some thoughts.

> 1. Let's change Role ids to just Role on the form.

Done!

> 2. I think it should be possible to bulk-invite fewer than three people on this screen. What if there are only one or two other people on my team?

I've been instructed to do three for now, and was told another developer on the team will work on a stimulus controller so we can adjust the amount of people we can invite for these invitations. If anything, I can just set it to one and make a note about the other developer adjusting the JS.

Also, concerning `roles_ids`, I was under the impression that we had to make this a single attribute when I received the instructions for the task, but now that I'm working through it I'd like to add `html_options: {multiple: true}` to the select element. In doing so we would leave the adding of roles up to the discretion of the developer, but it would work seamlessly with Membership creation and it also makes sense with how we currently set roles for single invitations via multiple checkboxes.

Also, will look into making the page with errors retain the values, not quite sure what the cause is right now.

> 3. As I noted in an inline comment, the test for not skipping this screen isn't being run in CI because the feature defaults to being turned off. I think we can manually set that config in the test setup, and then restore it during teardown.

I went ahead and edited the tests so that they run regardless of the setting in `config/initializers/bullet_train.rb`.

jagthedrummer:
@gazayas, thanks for the info about the JS portion of this. Are we planning to wait and have that be a part of this PR? Or are we wanting to merge this and then follow up later with the JS stuff? And, if we're planning to wait, is there some way to make the controller just be forgiving about one or more of those email addresses being empty?

I think making roles a multi select makes sense given that you can assign multiple roles in other places within the app. (Related, it might be nice to have a BT config option that toggles the `multiple` setting for roles across the whole app. (As a different issue/PR.))

jagthedrummer:
@gazayas, I just pulled down this branch and gave it another run and I think things are looking good for the portion that we're trying to implement in this PR.

Since we're not really considering this to be a completed feature (due to the JS portion for "add more team members") I think we should probably disable this bit from `bin/configure` for now. And then re-enable it when we ship the JS portion that will complete the feature.

https://github.com/bullet-train-co/bullet_train/pull/835/files#diff-1d9e0f07dcb2a76e98c85d6aa5c96b7bc60c14fd489570590950b54e0c35f9f4R177-R185

Other than than I think we can merge this once merge conflicts are resolved and tests are passing in both PRs.

gazayas:
@jagthedrummer Done!

jagthedrummer (approved):

PR #838, opened 2023-08-10
Try linking yarn in CI

jagthedrummer:
This makes it so that the `Core:` tests in CI will run against the JS provided by the gems/packages in `core` instead of running against the JS provided by the published versions on npm.

### DIFF
--- a/bin/checkout-and-link-core-repo
+++ b/bin/checkout-and-link-core-repo
@@ -58,3 +58,25 @@ updates="${packages[@]}"
 bundle lock --update $updates
 
 
+packages=(
+  "bullet_train"
+  "bullet_train-fields"
+  "bullet_train-sortable"
+)
+
+for package in "${packages[@]}"
+do
+  :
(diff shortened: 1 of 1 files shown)


pascallaliberte:
@jagthedrummer 
The best way to link packages is via the mechanisms already setup in ˋbin/hack`: we use the ˋyalc` utility after having compiled the js in a package.

From within the package’s root: `yarn build` and ˋyalc push`.

From within the main repo: `yalc link <packagename>ˋ

ˋyarn link` is prone to not properly reference sub-dependencies, while ˋyalc` creates a sort of local ˋnpm` repo and allows the inclusion of the local version of a package without modifying the package.json reference.

jagthedrummer:
The `Release: Minitest` CI step that's failing is because of a change in the starter repo that depends on changes in `core` and that test is running against the gems currently released to rubygems.

pascallaliberte (approved):

kaspth (commented):
nice!

PR #849, opened 2023-08-11
Themes localization - test suite upgrade

3pns:
This PR upgrades the test suites in regard to the related changes introduced in bullet_train-co https://github.com/bullet-train-co/bullet_train-core/pull/393

### DIFF
--- a/test/system/super_scaffolding_partial_test.rb
+++ b/test/system/super_scaffolding_partial_test.rb
@@ -113,10 +113,10 @@ class SuperScaffoldingSystemTest < ApplicationSystemTestCase
       find("#partial_test_multiple_buttons_test_two+button", visible: :all).click
       find("#partial_test_multiple_buttons_test_three+button", visible: :all).click
       # Date partial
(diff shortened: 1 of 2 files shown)


3pns:
It seems like the Core tests are picking up my branch in bullet_train-co but the Release tests are still running with the code from master as the same test fails on the Release steps. Also the error says it cannot find the element with the new id "#partial_test_date_test_display"

jagthedrummer:
Sorry I didn't mention this before, @3pns. The `Release` tests do indeed run against the version of the gems listed in `Gemfile` and `Gemfile.lock`. This is to help us know when we're introducing breaking changes. So I'd expect them to fail in this case, and it's not a blocker on merging the PR. It's just an indication that we need to release new versions of the `core` gems very soon after merging both PRs.

3pns:
@jagthedrummer Well noted. I ensured that Capybara can find the fields with both the old and new id so now the tests pass for release as well.

PR #899, opened 2023-08-23
Add configuration for incoming webhook parent class

gazayas:
Closes https://github.com/bullet-train-co/bullet_train-core/issues/454.

Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/461

We could technically put this in another file and write a block like this:

```ruby
BulletTrain::IncomingWebhooks::Configuration do |configure|
  ...
end
```

However, I think `BulletTrain::Configuration` is small enough right now so that we can put this all in one file.

I also used a String here and `constantize` it in the class itself to avoid constant loading issues on initialization.

### DIFF
--- a/config/initializers/bullet_train.rb
+++ b/config/initializers/bullet_train.rb
@@ -1,4 +1,7 @@
 BulletTrain.configure do |config|
   # Uncomment this line if you want to bypass strong passwords.
   # config.strong_passwords = false
+
+  # Change the parent class you want incoming webhooks to inherit from.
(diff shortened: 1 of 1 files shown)


gazayas:
Also, just to clear things up, the model `WorkspaceRecord` was mentioned in https://github.com/bullet-train-co/bullet_train-core/issues/454, but I'm assuming this is referring to a model that were only to exist in a downstream application, and isn't the default model we inherit from when running `bin/super-scaffold incoming-webhooks ProviderName`.

PR #962, opened 2023-09-04
Tests for `bin/hack --link`

gazayas:
This is a start for addressing https://github.com/bullet-train-co/bullet_train-core/issues/510.

This doesn't cover `bin/hack --link` for custom paths yet, but it covers the base logic for paths pointing to `local/bullet_train-core`.

@jagthedrummer I'll open this one as a draft in case we want to add custom paths in this PR or edit this base test in a way that would be the most productive/beneficial for the script as a whole.

### DIFF
--- a/test/scripts/hack_task_test.rb
+++ b/test/scripts/hack_task_test.rb
@@ -9,13 +9,13 @@ class HackTaskTest < ActiveSupport::TestCase
       line.match?('gem "bullet_train"')
     end
 
-    # Ensure the original bullet_train path stays in the temp Gemfile.
+    # Because these tests are being run against a task in the `bullet_train` gem,
+    # we ensure that the original `bullet_train` gem the developer is using stays in the temp Gemfile.
+    # We can afford to skip checking `path: ...` for this gem in the individual test cases
+    # because the `--link` logic is mostly covered for the other Bullet Train gems that are in the Gemfile,
+    # including ones that aren't originally there but are added with `bin/hack --link`.
     @temp_gemfile_lines = @temp_gemfile_lines.map do |line|
-      if line.match?('gem "bullet_train"')
-        @current_bullet_train_gem_version
-      else
-        line
-      end
+      line.match?('gem "bullet_train"') ? @current_bullet_train_gem_version : line
     end
 
     # Replace the Gemfile with the temp file.
@@ -24,9 +24,22 @@ class HackTaskTest < ActiveSupport::TestCase
 
   def teardown
     File.write("Gemfile", @original_gemfile_lines.join)
+    `bundle install`
   end
 
-  # TODO
-  test "bin/hack --link Links gems properly to local/bullet_train-core" do
+  test "bin/hack --link links gems properly to local/bullet_train-core" do
+    `bin/hack --link`
+
+    # Reload the Gemfile's contents.
+    edited_gemfile = File.open("Gemfile").readlines.join
+
+    framework_packages_path = `bundle show bullet_train`.chomp + "/config/locales/en/framework_packages.yml"
+    framework_packages_data = YAML.load(File.open(framework_packages_path))
+    framework_packages = framework_packages_data.dig("en", "framework_packages").map {|k, v| k}
+
+    framework_packages.each do |gem_name|
+      next if gem_name == "bullet_train"
+      assert edited_gemfile.include?("gem \"#{gem_name}\", path: \"local/bullet_train-core/#{gem_name}\"")
+    end
   end
 end
--- a/test/scripts/hack_task_test.rb
+++ b/test/scripts/hack_task_test.rb
@@ -35,7 +35,7 @@ class HackTaskTest < ActiveSupport::TestCase
 
     framework_packages_path = `bundle show bullet_train`.chomp + "/config/locales/en/framework_packages.yml"
     framework_packages_data = YAML.load(File.open(framework_packages_path))
-    framework_packages = framework_packages_data.dig("en", "framework_packages").map {|k, v| k}
+    framework_packages = framework_packages_data.dig("en", "framework_packages").map { |k, v| k }
 
     framework_packages.each do |gem_name|
       next if gem_name == "bullet_train"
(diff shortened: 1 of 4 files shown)


gazayas:
I wanted to originally load the packages with `YAML.load`, but they're not being properly loaded on CI for some reason. I resorted to writing in literal strings like we do with `bin/checkout-and-link-starter-repo`.

gazayas:
The HIDE_THINGS test is passing now, but that's mainly because I've added this conditional (CI links to `path: "./core/bullet_train-core"`):

```ruby
if Dir.exist?("local/bullet_train-core")
  test "bin/hack --link links gems properly to local/bullet_train-core" do
    `bin/hack --link`
    ...
  end
end
```

All of the tests are working locally though, even when running with both `CI=true` and `HIDE_THINGS=true`.

test/scripts/hack_task_test.rb
```diff
@@ -0,0 +1,68 @@
...
+    # we ensure that the original `bullet_train` gem the developer is using stays in the temp Gemfile.
+    # We can afford to skip checking `path: ...` for this gem in the individual test cases
+    # because the `--link` logic is mostly covered for the other Bullet Train gems that are in the Gemfile,
+    # including ones that aren't originally there but are added with `bin/hack --link`.
+    @temp_gemfile_lines = @temp_gemfile_lines.map do |line|
+      line.match?('gem "bullet_train"') ? @current_bullet_train_gem_version : line
+    end
+
+    # Replace the Gemfile with the temp file.
+    File.write("Gemfile", @temp_gemfile_lines.join)
```
jagthedrummer, line 47: I don't quite understand what this `setup` routine is supposed to do. If I insert a `binding.pry` at the end, after we're re-written `Gemfile` I'm not seeing that any actual change has been made. The line that says `gem "bullet_train"` prior to `setup` still says `gem "bullet_train"` after `setup` has run.

So I think that the main thing `setup` is doing currently is capturing the pre-test state of `Gemfile` so that we can restore it. If that's the case (and I'm not missing something) it might be simpler and easier to follow if we just explicitly made a copy of `Gemfile` and then restored it.

Running this test is also leaving my local repo in a dirty state, with uncommitted changes to `Gemfile.lock`, so we may need to restore that one as well.

Something like:

```ruby
def setup
  `cp Gemfile Gemfile.pre_test`
  `cp Gemfile.lock Gemfile.lock.pre_test`
end

def teardown
  `mv Gemfile.pre_test Gemfile`
  `mv Gemfile.lock.pre_test Gemfile.lock`
end
```

test/scripts/hack_task_test.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def teardown
+    File.write("Gemfile", @original_gemfile_lines.join)
+    `bundle install`
+  end
+
+  if Dir.exist?("local/bullet_train-core")
```
jagthedrummer, line 55: This line is currently making it so that we _don't_ run this test in CI since we don't have a `local/bullet_train-core` directory. In CI we currently clone the `core` repo into `./core` so it might make sense to pass that as the link destination.

test/scripts/hack_task_test.rb
jagthedrummer: Oh, I think maybe I understand why `setup` is the way it is.

In CI, due to the way `bin/checkout-and-link-core-repo` works the pre-test state of `Gemfile` would be saying:

```ruby
gem "bullet_train", path: "core/bullet_train"
```

Instead of just:

```ruby
gem "bullet_train"
```

And we'd want to keep pointing to `core/bullet_train` instead of `local/bullet_train-core/bullet_train` to ensure that we're testing against whichever branch of `core` we may have determined to use via our branch-matching process. Without that `setup` process we'd end up testing against the `main` branch, which is what we'd get by default when running `bin/hack --link`.

**BUT** since we don't _already_ have anything in `local/bullet_train-core` we end up skipping this test entirely, which is probably not what we want in CI.

I think we need to remove that conditional and try to figure out a different way around whatever is causing things to fail.

test/scripts/hack_task_test.rb
jagthedrummer: Scratch that. If we passed in `./core` as the link destination we'd expect the end result to be zero changes to the `Gemfile`, so we wouldn't be actually testing any functionality. 🤔

test/scripts/hack_task_test.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    "bullet_train-scope_questions",
+    "bullet_train-scope_validator",
+    "bullet_train-sortable",
+    "bullet_train-super_load_and_authorize_resource",
+    "bullet_train-super_scaffolding",
+    "bullet_train-themes",
+    "bullet_train-themes-light",
+    "bullet_train-themes-tailwind_css",
+  ]
```
jagthedrummer, line 27: Instead of maintaining this list it might suffice to just spot check two specific packages. For instance we could theoretically just test `bullet_train-api` as being representative of gems that _are_ listed in the `Gemfile` and test `bullet_train-fields` as being representative of gems that _are not_ listed in the `Gemfile`.

test/scripts/hack_task_test.rb
jagthedrummer: Just had a thought about this. I kinda think that we can safely assume that `bin/hack --link` should always be working against a "clean" `Gemfile`. That is, one that starts out pointing to released gems, and not to random local paths.

If we're comfortable with that assumption then can we just copy `test/support/temp_gemfile.rb` over the top of `Gemfile` in `setup` and then restore the original state in `teardown`?

```ruby
def setup
  `cp Gemfile Gemfile.pre_test`
  `cp Gemfile.lock Gemfile.lock.pre_test`
  `cp test/support/temp_gemfile.rb Gemfile`
  `bundle install`
end

def teardown
  `mv Gemfile.pre_test Gemfile`
  `mv Gemfile.lock.pre_test Gemfile.lock`
  `bundle install`
end
```

gazayas:
Didn't get around to this one today, will try to address it soon.

jagthedrummer:
No rush on this, imo, @gazayas. If there's other stuff that seems more urgent this can definitely wait.

PR #964, opened 2023-09-05
Run `bin/update/system_tests/use_device_test` in Bullet Train

kaspth:
Follow up to https://github.com/bullet-train-co/bullet_train/pull/720

### DIFF
--- a/test/system/account_test.rb
+++ b/test/system/account_test.rb
@@ -14,65 +14,59 @@ class AccountTest < ApplicationSystemTestCase
     end
   end
 
-  @@test_devices.each do |device_name, display_details|
-    test "user can edit their account on a #{device_name}" do
-      resize_for(display_details)
-      visit edit_account_user_path(@jane)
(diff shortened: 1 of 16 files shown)


test/system/account_test.rb
```diff
@@ -6,58 +6,50 @@ def setup
...
+    fill_in "Email", with: @jane.email
+    click_on "Next" if two_factor_authentication_enabled?
+    fill_in "Your Password", with: another_example_password
+    click_on "Sign In"
+    assert page.has_content?("Signed in successfully.")
+  end
 
-    test "user cannot edit password on a #{device_name} with invalid current password" do
-      resize_for(display_details)
-      login_as(@jane, scope: :user)
```
kaspth, line 45: Looks like when we delete `resize_for(display_details)` we're also deleting the next line, but we should only do that when the next line is a `\n`

jagthedrummer:
Oops, I forgot to point this at `main` before merging it. I'll open a new one.

PR #965, opened 2023-09-05
Use `build` in ExampleBot instead of `create` + `rollback`

jagthedrummer:
Joint PR: https://github.com/bullet-train-co/bullet_train-core/pull/524

This PR adds some dummy ids to some of the `*_example` factories. (But not any "live" factories that are used outside of `ExampleBot`.)

Doing this so that the values can be pre-populated for `ExampleBot` without us needing to actually create live records, clone them, rollback, then reset primary keys. Doing all of that work is slow so it's best to avoid doing it. (And having to reset primary keys for something that we expect to be a normal thing kind of gives me the willies.)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -28,7 +28,6 @@ Rails.application.configure do
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :memory_store
     config.public_file_server.headers = {
       "Cache-Control" => "public, max-age=#{2.days.to_i}"
     }
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -28,6 +28,7 @@ Rails.application.configure do
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
+    config.cache_store = :memory_store
     config.public_file_server.headers = {
       "Cache-Control" => "public, max-age=#{2.days.to_i}"
     }
--- a/test/factories/invitations.rb
+++ b/test/factories/invitations.rb
@@ -6,6 +6,8 @@ FactoryBot.define do
     sequence(:email) { |n| "example#{n}@email.com" }
 
     factory :invitation_example do
+      id { "42" }
+      from_membership_id { "42" }
       team { FactoryBot.example(:team) }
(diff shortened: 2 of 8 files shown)


newstler:
@jagthedrummer Well, I did some research, ran this on a real project, and to be honest I didn't find any noticeable differences besides empty timestamps which I fixed in last commit.  Sometimes I fell into `SystemStackError (stack level too deep)` with complicated includes of relations, but it appears that can be overcome with just setting `<relation>_id` instead of full relation example. I also changed ids from strings to integers which feels more natural for me to set, although strings were converted to integers anyway.

Not sure if ids should be hardcoded or sequenced like names so when there are list examples the ids would be different?

So basically I'm not sure what else can go wrong.
@andrewculver do you remember why we did `.create` instead of `.build` with examples generation? I remember we had a huge conversation and quite a strong base under this decision, but all I remember is that you worried that callbacks and such stuff would not fire, but I cannot really find anything that can cause problems in the code.

jagthedrummer:
@newstler, that sounds like good news! I don't have a strong opinion on generating ids/timestamps/etc vs hard coding them. If you prefer hard coding then lets do that. That's probably more stable since there's less that can unexpectedly change. I'll get these PRs rebased and cleaned up.

config/environments/development.rb
```diff
@@ -28,7 +28,6 @@
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :memory_store
```
jagthedrummer, line 31: @newstler I'm not sure we want to remove this. I think the point of this block is that you should be able to test a caching setup locally by having a file located at `tmp/caching-dev.txt`. If you don't want to test caching you'd just remove that file. Does that sound right to you?

newstler: Definitely that did it into PR occasionally, sure this line should stay unchanged.

jagthedrummer:
@newstler I just rebased the branches in both repos and gave everything one last review. Tests are all passing (other than when we test `core` against the released versions of the gems, and those failures are expected) and I think this is good to go. Do you have any other concerns before we merge this?

andrewculver:
Let's try it!

andrewculver (approved):

PR #967, opened 2023-09-06
Fix typo in bump-core banner

lagr:
This is just a tiny typo fix; I could not unsee it.


jagthedrummer (approved):
Thanks for the sharp eye, @lagr!

Issue #822, opened 2023-08-02
Should Gemfile.lock be removed from the starter repo?


jagthedrummer:
Every time that I pull updates from the starter repo `Gemfile.lock` is one of the messiest things to resolve. Often my local `Gemfile.lock` is already several version ahead of what the starter repo has.

I'm wondering if it would make sense to remove `Gemfile.lock` entirely from the starter repo. If we did we'd probably want to document the fact that people should commit the one that's generated for them to _their_ repo.

Possible complication: This might cause problems for people who already have a BT app going because when they pulled from the starter repo it would try to delete their `Gemfile.lock`. But we could probably document that in the upgrade docs.

gazayas:
I personally remember having a really hard time with https://github.com/bullet-train-co/bullet_train-core/issues/121 because the starter repository didn't define devise's version, so when running CI it got a different result from what I was getting locally, and that caused me a lot of headaches.

Also reminds me of this one: https://discord.com/channels/836637622432170028/966439333827592212/1081058807221862470

I can understand how this can be messy to resolve when pulling updates though, so I'm not quite sure what the best answer is here.

olbrich:
If you try to run `bundle install` on a conflicted lock file it will tell you to..
	
	git checkout HEAD -- Gemfile.lock
	bundle install 

Which generally does the trick for me. If the Gemfile has appropriate version constraints on gems, then this should be all that's needed, right?

jagthedrummer:
Unfortunately the `Gemfile` has almost zero constraints on gems. But I would _much_ prefer it if it did.

jagthedrummer:
We've started to pin the version number for BT gems, and have documented in the upgrade guide how to deal with messy conflicts in `Gemfile.lock`. I think that's good enough for now, so I'm closing this issue. Though in the future we may want to move towards pinning explicit version numbers for even more gems in `Gemfile`.

PR #968, opened 2023-09-06
Explicitly set the expected verison of BulletTrain gems

jagthedrummer:
Since the starter repo is so closely tied to all of the gems in the `core` repo we need to be more explicit about which version we expect to have.

Joint PR: https://github.com/bullet-train-co/bullet_train-core/pull/529

Related to #822 

TODO:

- [x] Update `Gemfile` to explicitly set versions
- [x] Update `package.json` to explicitly set versions
- [x] Update the "generate version bump PR" GitHub action to be able to set the version in `Gemfile` and `package.json` (Example PR that downgrades all packages by one patch level https://github.com/bullet-train-co/bullet_train/pull/971)
- [x] Update `bin/hack` https://github.com/bullet-train-co/bullet_train-core/pull/529
- [x] Add the `core` gems that are not already in `Gemfile` to `Gemfile`? (I think we need to do this to be able to pin their versions explicitly. Or we have to update the version bump actions in `core` to explicitly alter the `gemspec` of the gems that require the unlisted gems as dependencies.)
- [x] Update upgrade docs - https://github.com/bullet-train-co/bullet_train-core/pull/532


kaspth:
> Add the core gems that are not already in Gemfile to Gemfile? (I think we need to do this to be able to pin their versions explicitly. Or we have to update the version bump actions in core to explicitly alter the gemspec of the gems that require the unlisted gems as dependencies.)

Good call on this! I think hoisting them from being transient dependencies to being in the Gemfile, is the best option.

kaspth (approved):

pascallaliberte (approved):

Gemfile
```diff
@@ -100,21 +100,35 @@ end
 # BULLET TRAIN GEMS
 # This section is the list of Ruby gems included by default for Bullet Train.
 
+# We use a constant here so that we can ensure that all of the bullet_train-*
+# packages are on the same version.
+BULLET_TRAIN_VERSION = "1.3.24"
```
gazayas, line 105: Love this 💯
Will be easy for developers to spot and change too.

gazayas (approved):

Gemfile
olbrich: This is how I do it.

Gemfile
```diff
@@ -100,21 +100,35 @@ end
...
-gem "bullet_train-scope_questions"
-gem "bullet_train-obfuscates_id"
+gem "bullet_train-sortable", BULLET_TRAIN_VERSION
+gem "bullet_train-scope_questions", BULLET_TRAIN_VERSION
+gem "bullet_train-obfuscates_id", BULLET_TRAIN_VERSION
+
+# Core gems that are dependencies of gems listed above. Technically they
+# shouldn't need to be listed here, but we list them so that we can keep
+# verion numbers in sync.
+gem "bullet_train-fields", BULLET_TRAIN_VERSION
```
olbrich, line 126: Can't the gems that depend on them specify the version in their gemspec?

kaspth: It seems safer that it's in the Gemfile rather than locked in a gemspec. In case something goes wrong and people need a fix where they can't downgrade, they'd need to fork and revise the gemspec version.

Of course, that's hypothetical and I don't know if we'd ever need this. Having the lever seems better to me though.

olbrich: I'm not positive if this would work, but I think if you explicitly set a version in the Gemfile it may override the version defined in the gemspec. That way you could downgrade if you really needed to.

olbrich: Just a thought...
changing this line

https://github.com/bullet-train-co/bullet_train-core/blob/13368d59430d45d25a907f79ad3b29bd865f6b60/bullet_train/bullet_train.gemspec#L36

to something like...

```
  spec.add_dependency "bullet_train-fields", BulletTrain::VERSION
```

Might help.  This would ensure that the bullet_train gem depends on the same version of the related gems. These are all now being published in lock-step now, so I don't think it would be a problem.

jagthedrummer: I don't think that there's a way to do that in the various `.gemspec` files that wouldn't result in the problem that @kaspth described.

Like if for some (extremely weird) reason someone wanted to be on version `1.3.25` for _most_ of the BT gems, but wanted to be on `1.3.24` for `bullet_train-fields` they'd have no easy way to do that.

In order to keep things in sync `bullet_train` would have to declare a hard dependency on `bullet_train-fields` for _exactly_ version `1.3.25` and if you asked for `1.3.24` of `bullet_train-fields` in your `Gemfile` then `bundler` would complain at you about incompatible versions.

I think we want to leave the version requirements in `bullet_train.gemspec` as loose as possible and leave it up to the "consuming app" (aka the starter repo) to manage explicit version dependencies.

PR #985, opened 2023-09-11
Super Scaffold and test an incoming webhook

gazayas:
Solves the remaining problem in https://github.com/bullet-train-co/bullet_train-core/issues/453.

Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/543

We aren't running a system test, but by Super Scaffolding an incoming webhook and testing it this way, we can ensure that the scaffolder itself works.


gazayas:
The release test is failing as expected, but the core Super Scaffolding test is passing so we should be good to go.

Issue #1007, opened 2023-09-18
dates_helper_test makes assumptions about root_path


olbrich:
https://github.com/bullet-train-co/bullet_train/blob/742ba58bf5f73666e69b744e4532d3396238a5bb/test/system/dates_helper_test.rb#L66

This test assumes that the root_path will result in the account dashboard, which is not always true.

I suggest changing this to 
```
        visit  account_team_scaffolding_absolutely_abstract_creative_concepts_path(Team.find_by!(name: "The Testing Team"))
```

PR #1018, opened 2023-09-20
Write more precise routing in system tests

### LINKED ISSUE #990: Invitation tests use wrong route
https://github.com/bullet-train-co/bullet_train/blob/40a29c70340b1642080d095a53f5da32bc856c30/test/system/invitations_test.rb#L287

This should be the new_user_session_path route. It fails if your root_path doesn't redirect to a login when the user is logged out.

gazayas:
Opening this in an effort to close #990 and #1007.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -153,7 +153,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     assert_text "Signed out successfully"
   end
 
-  def sign_in_from_homepage_for(display_details)
+  def new_session_page_for(display_details)
     # We actually want to go straight to the user session and
     # not the home page in case the home page doesn't have new session logic.
     visit new_user_session_path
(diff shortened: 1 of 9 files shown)


test/application_system_test_case.rb
```diff
@@ -150,18 +150,19 @@ def sign_out_for(display_details)
...
-    # TODO the tailwind port of bullet train doesn't currently support a homepage.
+    # We actually want to go straight to the user session and
+    # not the home page in case the home page doesn't have new session logic.
     visit new_user_session_path
 
     # this forces capybara to wait until the proper page loads.
     # otherwise our tests will immediately start trying to match things before the page even loads.
     assert_text("Sign In")
   end
+  alias_method :new_session_page_for, :sign_in_from_homepage_for
```
gazayas, line 165: I added the alias method here and the other below because I think it makes more sense naming them this way. I also used an alias method instead of simply re-writing the method in case any developers decided to use this method for their own tests.

gazayas:
@jagthedrummer This one's ready to go. I followed the setup directions here to use another root page:

https://github.com/bullet-train-co/bullet_train/blob/daf99d07a23452515ddaa795f45d27a63e9ecbe1/app/controllers/public/home_controller.rb#L3-L4

test/application_system_test_case.rb
jagthedrummer: I like the new names. I wonder if we should use those names for the actual methods and then use the aliases to preserve backwards compatibility (for now). Or maybe instead of just a raw `alias_method` we implement an actual method that would show a deprecation warning and then call the original method?

Something like:

```ruby
def sign_in_from_homepage_for(display_details)
  puts "sign_in_from_homepage_for is deprecated".red
  puts "  please switch to: new_session_page_for".red
  puts "  called from #{caller.first}".red
  new_session_page_for(display_details)
end
```

jagthedrummer (approved):
@gazayas, this looks great! I opened https://github.com/bullet-train-co/bullet_train/pull/1024 (which targets _this_ PR branch) as a proposal for taking this a little further. Let me know what you think. If you're not in favor we can merge this on as-is.

PR #1024, opened 2023-09-21
Switch all tests to new names and provide deprecation warnings about old names

jagthedrummer:
A proposal for taking https://github.com/bullet-train-co/bullet_train/pull/1018 a little bit further. (Note that this PR targets the branch for that PR, and not `main` directly.)


gazayas:
@jagthedrummer Yup, I'm all about it, I prefer these method names 🙌

PR #1027, opened 2023-09-22
Add last team exception and team system test for Membership foreign constraint

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/575


config/initializers/exceptions.rb
```diff
@@ -1 +1,3 @@
 class RemovingLastTeamAdminException < RuntimeError; end
+
+class RemovingLastTeamException < RuntimeError; end
```
jagthedrummer, line 3: Doesn't need to be a holdup on this PR, but it seems like these errors should live in the `bullet_train` gem since that's where they're raised and rescued.

PR #1030, opened 2023-09-22
Update Super Scaffolding system test setup for automated `rails g` commands

gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/547


gazayas:
I made some progress on this one. The remaining models that are commented out in the Super Scaffolding system test setup script have the same namespace problem that I handled in https://github.com/bullet-train-co/bullet_train-core/pull/503. I solved it with the following code.

https://github.com/bullet-train-co/bullet_train-core/blob/main/bullet_train-super_scaffolding/lib/bullet_train/super_scaffolding/scaffolders/crud_scaffolder.rb#L40-L52

I'm hoping to implement something similar to account for the namespaces in the system test here, because you can see by the following commands that `references` attribute name doesn't use the entire parent name.

```
bundle exec spring rails g model Personality::Note disposition:references name:string
bin/super-scaffold crud Personality::Note Personality::Disposition,Team name:text_field description:trix_editor
```

Issue #1028, opened 2023-09-22
InvitationListsTest is flaky


jagthedrummer:
@gazayas, I'm not sure why, but the new `InvitationListsTest` seems to be pretty flaky in CI. We're getting a bunch of random failures that go away when they're re-run.

Unfortunately it doesn't seem to be the same error every time. 😞 

Here are a few examples.

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2497/workflows/d64d641f-0405-4639-9099-7f2ab3153922/jobs/12714/parallel-runs/1/steps/1-114?invite=true#step-114-19190_88

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2494/workflows/1b5b141c-03e2-4756-802c-922b1ea39621/jobs/12699/parallel-runs/0/steps/0-113?invite=true#step-113-2434_89

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2486/workflows/632debd0-899f-4c9a-b585-9213f02f649f/jobs/12637/parallel-runs/1/steps/1-112?invite=true#step-112-12271_88

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2481/workflows/f35af605-b1da-4ae1-a48a-2f7f43842eaa/jobs/12602/parallel-runs/0/steps/0-114?invite=true#step-114-13762_89

jagthedrummer:
I think this was fixed in https://github.com/bullet-train-co/bullet_train/pull/1034

jagthedrummer:
It looks like maybe this wasn't fixed. Here's a failure in that test:

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2586/workflows/563d4134-34be-4e1b-b7fb-35627f6cf850/jobs/13424

And after re-running it passed:

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2586/workflows/f5751b56-67c8-4a77-bd21-fe41797781e2/jobs/13430

jagthedrummer:
Another instance:

https://app.circleci.com/pipelines/github/bullet-train-co/bullet_train/2654/workflows/51c09053-5dac-4ece-b851-2cd416dabd46/jobs/13994?invite=true#step-113-6731_163

jagthedrummer:
Still happening:

[...]

PR #1034, opened 2023-09-25
Add temporary fix to enable headless invitation lists tests

### LINKED ISSUE #1029: System tests have started always opening a visible browser
Recently when I run system tests locally it's always opening a visible browser.

Until recently system tests would always run invisibly in the background unless I manually added `MAGIC_TEST=1` to the test command.

I've verified that I don't have `MAGIC_TEST` set in my environment.

```
$ echo $MAGIC_TEST

$
```

gazayas:
Fixes #1029.

For some reason in the Invitation List tests only, we get the following value.
```
     5: def setup
 =>  6:   binding.pry
     7: 
     8:   @current_bulk_invitations_setting = nil
     9:   BulletTrain.configure do |config|
    10:     @current_bulk_invitations_setting = config.enable_bulk_invitations
    11:     config.enable_bulk_invitations = true
    12:   end
    13: end

[1] pry(#<InvitationListsTest>)> Capybara.current_driver
=> :selenium
```

However, in [application_system_test_case.rb](https://github.com/bullet-train-co/bullet_train/blob/main/test/application_system_test_case.rb), we only set one of these two keys explicitly.

https://github.com/bullet-train-co/bullet_train/blob/d95deaa09ed30441cd4c318bdff315e73a5c6a5a/test/application_system_test_case.rb#L26-L27

I'm guessing this is related to #1028 especially since it's a Selenium::Webdriver error that's being raised, but I'll have to investigate further to make sure.

### DIFF
--- a/test/system/invitation_lists_test.rb
+++ b/test/system/invitation_lists_test.rb
@@ -3,6 +3,10 @@ require "application_system_test_case"
 class InvitationListsTest < ApplicationSystemTestCase
   @@test_devices.each do |device_name, display_details|
     def setup
+      # TODO: Capybara.current_driver is returning :selenium,
+      # when it should be either :selenium_chrome or :selenium_chrome_headless.
+      Capybara.current_driver = Capybara.default_driver
+
       @current_bulk_invitations_setting = nil
       BulletTrain.configure do |config|
         @current_bulk_invitations_setting = config.enable_bulk_invitations


gazayas:
Passing consistently now

スクリーンショット 2023-09-25 19 49 22

jagthedrummer (approved):
Wow, good find! I don't think it would have even occurred to me to check that.

PR #1070, opened 2023-10-10
Drop old unused tables

jagthedrummer:
These two tables seem to be remnants of something that's been removed.


jagthedrummer:
@andrewculver gave me a gut check on this via Discord and said that this PR is good to go.

Issue #978, opened 2023-09-08
Can we provide a GitHub action that automatically generates a PR to upgrade an app?


jagthedrummer:
@kaspth brought this up in Discord, and I think it's a great idea and certainly worth exploring.

One challenge is how we handle merge conflicts. The process would generate a PR to perform the upgrade, instead of performing the upgrade directly. So we think that if there are merge conflicts we could clobber whatever changes have been made in the downstream app, and have the PR show the changes from the starter repo. We think that would effectively move when conflict resolution takes place. In the normal manual process conflict resolution happens at the time that the changes from the starter repo are merged into the upgrade branch. We think that if we always pick the upstream changes from the starter repo that it will just move conflict resolution to happen at the time that the upgrade PR is reviewed/merged.

Some (possibly pseudo-code) bash to illustrate the idea:

```bash
# !/usr/bin/env bash
# bin/upgrade

set -ex

git fetch bullet-train
git fetch --tags bullet-train

tag=$(bundle show bullet_train) # Find the version?

# Figure out how to bump the version

git checkout -b updating-bullet-train-${tag}
git merge $tag -s=theirs

conflicts=$(git update-index --refresh && git diff-index --quiet HEAD --) # Not sure if this command is right?

repo_url=$(git remote show origin) # Somehow get the url?

if test [ -z $conflicts ]; then
  # Open an issue with the conflicts, describing to run `bin/upgrade` locally.
  curl -POST $repo_url/"api/v3/issues"
else
  # Open a PR that will hopefully pass and can be merged.
  curl -POST $repo_url/"api/v3/pulls"
fi
```

jagthedrummer:
Relatedly, @olbrich [suggested](https://github.com/bullet-train-co/bullet_train-core/pull/532#discussion_r1320148226) that `bootboot` from Shopify might be an interesting way to help smooth the upgrade process. I don't know if that would be helpful here. https://github.com/Shopify/bootboot

kaspth:
Since we're doing things similar to dependabot there may be some inspiration to take from @marcoroth's https://github.com/marcoroth/dependabot-bump-together-action too

PR #1077, opened 2023-10-12
Add GitHub workflow for updating the Bullet Train version

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train/issues/978

This is my first stab at making a workflow that will generate a PR for updating an app against the Bullet Train starter repo.

## Limitations

In order to generate the PR we're using the `theirs` resolution method to deal with any conflicts. In this context that means that if a block of code has been changed in the downstream app _and_ in the starter repo, then the PR that we generate will contain a change set that will clobber the changes in the downstream repo in favor of the change that was made in the starter repo.

For instance, my app was on `1.5.2`, and I changed this line:

https://github.com/bullet-train-co/bullet_train/blob/2ea5c9c136568bae39257af6c8944267b058dc17/test/system/invitation_lists_test.rb#L29

to read:

```ruby
assert page.has_content?("Tell")
```

And then when I ran this action in my project to get to `1.6.0` the resulting PR proposed this change which clobbered my test change:

```diff
- assert page.has_content?("Tell")
+ assert_text("Tell us about you")
```

That change was just buried in amongst all the other changes in the PR and would be very easy to miss during a review.

This seems like it might be a larger footgun than we want to ship with the project.

I haven't been able to figure out how to make it so that GitHub would allow the PR to contain the proposed change, but for it to be seen as a merge conflict that would have to be sorted out.

We could potentially **not** attempt to resolve conflicts in any way, and just commit conflicted files into the PR branch.

In that case the PR would contain this:

```
<<<<<<< HEAD
    assert page.has_content?("Tell")
||||||| 2ea5c9c
    assert page.has_content?("Tell us about you")
=======
    assert_text("Tell us about you")
>>>>>>> v1.6.0
```

And then it would be up to the developer to spot that during review and to sort it out. Or maybe we could detect that there are conflicts and then flag the resulting PR in some way.

I'm not really sure... Would love to hear some thoughts on this.

## Requirements

Currently this workflow requires that a Personal Access Token be created and set in the repo secrets as `PAT`.

I opted to use a PAT because of limitations on the default token that is used for GitHub actions. Any workflows using the default token can't trigger new workflows as side effects. Specifically for this action it means that CI workflows for the resulting PR are not run when using the default token.

The token can be a fine-grained token as long as it has "Contets" and "Pull Request" permissions set to read/write.

Screenshot 2023-10-11 at 14-44-57 Fine-grained Personal Access Tokens

## Real-world testing

I've used this workflow to update my Seshy application from `1.5.0` through each step to `1.6.2`.

* 1.5.1
* 1.5.2
* 1.6.0
* 1.6.1
* 1.6.2

For each version I also manually prepared a PR by following the upgrade docs and manually sorting out conflicts in `Gemfile.lock`.

In each case the PR generated by this workflow was the same as the one that I generated manually. (Though for most of these version bumps the majority of changes were directly related to bumping the version number of BT packages, and they didn't contain a lot of _other_ changes to "normal" parts of the app.)


.github/workflows/update-bullet-train.yml
```diff
@@ -0,0 +1,83 @@
...
+        description: 'Version Number'
+        required: true
+        type: string
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
kaspth, line 18: There's a newer version of this package, might as well 😄

```suggestion
        uses: actions/checkout@v4
```

kaspth (approved):
Really cool! 🚀

kaspth:
> And then it would be up to the developer to spot that during review and to sort it out. Or maybe we could detect that there are conflicts and then flag the resulting PR in some way.

I wonder if we could use GitHub's Checks API to highlight this? https://github.blog/2018-05-07-introducing-checks-api/
Here's a sample of a thing that uses it https://github.com/andrewmcodes/standardrb-action/blob/65523383ea16a83a32a4226e8f81e91ae6160be7/lib/github_check_run_service.rb — though the README notes that there's a limitation on 50 notes per run.

pascallaliberte (approved):
Neat 👍

kaspth:
@jagthedrummer oh yeah, sorry, I didn't elaborate on this enough. I think your idea with leaving conflicts in there is good — and then maybe we could use the checks API to highlight them on GitHub and prevent a merge? 

But actually, we could probably open the PR as a draft PR in case we can see merge conflicts to help indicate that user action is required?

gazayas:
Just looked at the yml file, this is pretty awesome!

As for the conflicts, I definitely think we'd be shooting ourselves in the foot if we tried to resolve everything automatically, at least initially since we're first diving into doing automated app updates and I'm not sure what issues we might run into.

However, I'm all for showing the conflicts in a PR like you mentioned:

```
<<<<<<< HEAD
    assert page.has_content?("Tell")
||||||| 2ea5c9c
    assert page.has_content?("Tell us about you")
=======
    assert_text("Tell us about you")
>>>>>>> v1.6.0
```

I think this is in and of itself a productivity boost in that a developer could just pull down the branch and resolve things there without thinking much about the documentation.

## Ejected Partials
Because of https://github.com/bullet-train-co/bullet_train-core/pull/558, I think we'll need to check for ejected themes and grab the conflicts accordingly. Maybe we could:
1. Run a bash script that finds all folders under `app/views/themes`.
2. Get the name for each one and eject the themes with `rake bullet_train:themes:light:ejected[theme_name]`
3. Remove any extra files that weren't there in the downstream app originally, and then check for conflicts.

jagthedrummer:
@gazayas I like the idea for trying to deal with ejected themes, and it would be great to extend that to _any_ files that have been ejected if we can. But let's save that for a follow up effort.

For now I'm going to update this to _not_ try to resolve any conflicts other than in `Gemfile.lock`. And if we detect that there are conflicts it'll mark the PR in some way, or open it as a draft.

Thanks for the feedback, y'all!

jagthedrummer:
OK @kaspth, @gazayas, and @pascallaliberte. I've made most of the changes suggested above. I haven't quite figured out how we'd use the Checks API, so I'm skipping that for now.

I do have it set up so that it **will not** try to resolve any conflicts (other than in `Gemfile.lock`). If it detects that there _are_ conflicts it will mention that fact in the PR body, along with a list of conflicted files, and some steps to use for getting started on resolving them.

Here's an example PR body that I generated taking my app from 1.6.4 to 1.6.5:

---

Update of the Bullet Train Starter Repo to version `1.6.5`

**Please note:** You should thorougly review this PR.

⚠️  **There are conflicts in the following files:**
* test/test_helper.rb
 
**You should pull down this branch, resolve the conflicts, and then push the branch.**
 
Get started by pulling the branch:
```
git fetch origin
git checkout -b bullet-train-update/1.6.5 --track origin/bullet-train-update/1.6.5
```
Then resolve conflicts. Once you have done that you can commit the changes and then push.
```
git add .
git commit -m "resolved conflicts"
git push origin bullet-train-update/1.6.5
```

> Auto-generated by `.github/workflows/update-bullet-train.yml`

jagthedrummer:
I wonder if we should use [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to get most of this code out of the starter repo. 🤔 

If we did then this workflow file could be something like:

```yaml
name: "Create Bullet Train Update PR"
on:
  workflow_dispatch:
    inputs:
      versionNumber:
        description: 'Version Number (latest if blank)'
        type: string

jobs:
  update:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v4
        with:
          # We have to fetch the full history so that the merge step will understand
          # how to properly merge the starter repo with our local changes.
          fetch-depth: 0

      - name: Create update PR
        uses: bullet-train-co/create-update-pr@v1
```

jagthedrummer:
I just ran a quick test on the composite action idea, and with [this tiny repo](https://github.com/bullet-train-co/create-upgrade-pr) supplying a composite action I could get this workflow file down to a svelte 17 lines.

```yaml
name: "Create Bullet Train Upgrade PR"
on:
  workflow_dispatch:
    inputs:
      versionNumber:
        description: 'Version Number (latest if blank)'
        type: string

jobs:
  update:
    runs-on: ubuntu-latest
    steps:
      - name: Create upgrade PR
        uses: bullet-train-co/create-upgrade-pr@v1
        with:
          versionNumber: ${{ inputs.versionNumber }}
          githubToken: ${{ secrets.PAT }}
```

I'm inclined to go with this arrangement in keeping with the general Bullet Train ethos of providing functionality without the code for that functionality needing to live directly in your application repo.

jagthedrummer:
Thanks, @kaspth! Excellent idea for this on your part. It was kind of a fun one to figure out. I hope that it works for other people's apps as well. If it does work out we might investigate how to get it to run on a schedule so that you magically get an upgrade PR whenever you need it. Or maybe it would have to be a GitHub app that you add to your repo? 🤔

kaspth:
>  If it does work out we might investigate how to get it to run on a schedule so that you magically get an upgrade PR whenever you need it. Or maybe it would have to be a GitHub app that you add to your repo? 🤔

Maybe we could start with just a daily `schedule:` to run the script? I don't know what GitHub apps are or how they work, so I'm not sure about that.

PR #1078, opened 2023-10-12
Remove testing for use_device_test transition script

jagthedrummer:
In #720 we added a bit of setup to the CircleCI tests to run the `use_device_test` script in CI and then re-run the tests.

Now that we've merged that PR and the follow-on PR that changed the format of all the system tests, we no longer need to run that script and re-test in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,16 +125,6 @@ jobs:
           environment:
             KNAPSACK_PRO_CI_NODE_TOTAL: 2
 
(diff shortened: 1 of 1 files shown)


jagthedrummer:
@kaspth, is there any reason that we should keep this?

kaspth:
@jagthedrummer no, I forgot to reapply my commit that removed it after one of the rebases. We should remove this!

kaspth (approved):

Issue #1094, opened 2023-10-16
Lingering use of `resize_for`


jagthedrummer:
@kaspth, I just noticed that we still have a test using `resize_for`.

```
$ rails test test/system/animations_test.rb
...

DEPRECATION WARNING: `resize_for` is deprecated.
Please run the following command to update your tests:
./bin/updates/system_tests/use_device_test
 (called from block in <class:AnimationsTest> at /Users/jgreen/projects/bullet-train-co/bullet_train/test/system/animations_test.rb:17)

...
```

Is it right that all we need to do is to switch that to `resize_display`?

PR #1095, opened 2023-10-16
Replace a lingering instance of `resize_for`

jagthedrummer:
We cleaned up most of these in
https://github.com/bullet-train-co/bullet_train/pull/720 but the conversion script couldn't see this one.

Fixes https://github.com/bullet-train-co/bullet_train/issues/1094

### DIFF
--- a/test/system/animations_test.rb
+++ b/test/system/animations_test.rb
@@ -14,7 +14,7 @@ class AnimationsTest < ApplicationSystemTestCase
 
     display_details = {resolution: [750, 1334], mobile: true, high_dpi: true}
     animation_duration = 0.5
-    resize_for(display_details)
+    resize_display(display_details)
     login_as(@jane, scope: :user)
     visit root_path
     if billing_enabled?


kaspth (approved):
Nice! Now that we've got the tests updated, I wonder if we should change the deprecation message to say `use resize_display` instead?

jagthedrummer:
> Now that we've got the tests updated, I wonder if we should change the deprecation message to say `use resize_display` instead?

Hmmm.... I'm really not sure about that.

If people have been keeping up on versions that could make sense, but I'm concerned about what happens to people that haven't been keeping up.

Like imagine someone is on `1.2.x` (or earlier) and has written their own system tests that were using the old pattern that we replaced with `device_test`. They'd have some instances where the existing deprecation message is telling them to do the right thing.

Maybe we should change the message to say that they could _try_ running the helper to clean it up, and we could also include a link to a page with more details about the change and how to manually fix any lingering references?

kaspth:
> Maybe we should change the message to say that they could try running the helper to clean it up, and we could also include a link to a page with more details about the change and how to manually fix any lingering references?

This sounds good to me!

Issue #764, opened 2023-05-04
Testing workflow for GitHub


jagthedrummer:
It would be nice to have a working GitHub workflow included with the starter repo.

Circle is OK, but since all of the docs are built around an assumption of using GitHub it would be super convenient to have a testing pipeline working out of the box with not additional configuration or services required.

gazayas:
Linking https://github.com/bullet-train-co/bullet_train/pull/186 here

Not sure about the progress there, but it's definitely been considered.

Issue #1010, opened 2023-09-19
Github action improvements


olbrich:
@jagthedrummer Some suggestions for Github actions that may be useful to downstream apps.

1. Disable the `tag & release` and `version bump` workflows when they are not being run in the starter app repo.  Downstream users can always copy them and enable them for their own purposes but right now they are fairly BT specific.

2. Provide a template for a GH action that will run tests and system tests.  It may be disabled / commented out by default, but it would be helpful for users that don't also use CircleCI.

jagthedrummer:
We added GitHub Actions workflows in https://github.com/bullet-train-co/bullet_train/pull/1106 (and switched to using them for our internal CI purposes).

The `tag & release` and `version bump` workflows are definitely BT specific, but I'm not sure there's a great way to disable them while leaving them in place.

We could potentially add a step to `bin/configure` to remove them completely, but then I think if we made changes to those files in the future it would make for a weird merge when people upgraded their BT version.

I think it may be better to just rely on people removing them from their own copy of the app if they don't want to see the hanging around.

Gonna close this issue.

PR #1106, opened 2023-10-20
Add CI/CD workflows for GitHub Actions

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train/issues/764
Partially addresses https://github.com/bullet-train-co/bullet_train/issues/1010

This adds several new workflows for GitHub actions (and slightly tweaks some existing workflows to get them to display nicer with the new ones).

CleanShot 2023-10-20 at 13 04 28

The `💻 PR CI/CD Pipeline (staging)` workflow is configured to automatically run whenever a PR is opened. It will run the main test suite, the super scaffolding tests, standardrb, and a database schema check. Tests are run in a matrix to speed things up, and the matrix can be configured in the workflow. The workflow can also be configured to auto-deploy to a staging app once the tests pass.

CleanShot 2023-10-20 at 13 18 51

The `🎥 Main CI/CD Pipeline (prod)` is configured to automatically run whenever a commit lands on `main` either via a PR being merged or from a direct push. It does all the same stuff as the PR workflow, and it can also be configured to auto deploy to a production app once tests pass.

CleanShot 2023-10-20 at 13 24 12

### Note for internal Bullet Train team

I think these pipelines should also be useful for the starter repo itself and could potentially be a stepping stone to moving stuff from CircleCI to GitHub actions. As they stand now I think these pipelines would replace 4 of the 8 jobs that we run in CircleCI.

CleanShot 2023-10-20 at 13 31 21

### DIFF
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -1,4 +1,4 @@
-1.2.13# This workflow will ensure that when we run `rails db:migrate` that no changes are made to `db/schema.rb`.
+# This workflow will ensure that when we run `rails db:migrate` that no changes are made to `db/schema.rb`.
 #
 # This workflow is pimarily meant to be called by other workflows, but it can be run manually.
 name: "🔎 ~ Database Schema Check"
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -1,0 +1,27 @@
+# This workflow will ensure that when we run `rails db:migrate` that no changes are made to `db/schema.rb`.
+#
+# This workflow is pimarily meant to be called by other workflows, but it can be run manually.
+name: "🔎 ~ Database Schema Check"
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    env:
+      RAILS_ENV: test
+      BUNDLE_JOBS: 2
+      BUNDLE_RETRY: 3
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Ruby and gems
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+
+      - name: Database Schema Check
+        id: db-schema-check
+        run : bash ./.circleci/db_schema_check
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -26,6 +26,7 @@ jobs:
           --health-retries 5
     env:
       RAILS_ENV: test
+      DATABASE_URL: "postgres://rails:password@localhost:5432/rails_test"
       BUNDLE_JOBS: 2
       BUNDLE_RETRY: 3
     steps:
@@ -36,9 +37,6 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
-
-      - name: Set up database schema
-        run: bin/rails db:schema:load
 
       - name: Database Schema Check
         id: db-schema-check
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -37,7 +37,10 @@ jobs:
         with:
           bundler-cache: true
 
+      - name: Set up database schema
+        run: bin/rails db:schema:load
+
       - name: Database Schema Check
         id: db-schema-check
-        run : bash ./.circleci/db_schema_check
+        run : ./.circleci/db_schema_check
         continue-on-error: false
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -1,4 +1,4 @@
-# This workflow will ensure that when we run `rails db:migrate` that no changes are made to `db/schema.rb`.
+1.2.13# This workflow will ensure that when we run `rails db:migrate` that no changes are made to `db/schema.rb`.
 #
 # This workflow is pimarily meant to be called by other workflows, but it can be run manually.
 name: "🔎 ~ Database Schema Check"
@@ -9,6 +9,21 @@ on:
 jobs:
   check:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11-alpine
+        ports:
+          - "5432:5432"
+        env:
+          POSTGRES_DB: rails_test
+          POSTGRES_USER: rails
+          POSTGRES_PASSWORD: password
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     env:
       RAILS_ENV: test
       BUNDLE_JOBS: 2
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -40,3 +40,4 @@ jobs:
       - name: Database Schema Check
         id: db-schema-check
         run : bash ./.circleci/db_schema_check
+        continue-on-error: false
--- a/.github/workflows/_reusable_workflows.yml
+++ b/.github/workflows/_reusable_workflows.yml
@@ -8,5 +8,4 @@ jobs:
   nothing:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
+      - run: echo "Nothing to see here. Move along. 🎇"
--- a/.github/workflows/_reusable_workflows.yml
+++ b/.github/workflows/_reusable_workflows.yml
@@ -1,6 +1,6 @@
 # This is a dummy workflow that's just here to add a bit of organization to the workflow list.
 # Too bad they don't allow you to hide workflows or something.
-name: " 🚫 |----- Reusable Worflows Below"
+name: " 🚫 |----- Reusable Worfklows Below"
 on:
   workflow_call:
 
--- a/.github/workflows/_reusable_workflows.yml
+++ b/.github/workflows/_reusable_workflows.yml
@@ -1,0 +1,12 @@
+# This is a dummy workflow that's just here to add a bit of organization to the workflow list.
+# Too bad they don't allow you to hide workflows or something.
+name: " 🚫 |----- Reusable Worflows Below"
+on:
+  workflow_call:
+
+jobs:
+  nothing:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
--- a/.github/workflows/_run_super_scaffolding_tests.yml
+++ b/.github/workflows/_run_super_scaffolding_tests.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         # For super scaffolding tests we need to have exactly 7 runners.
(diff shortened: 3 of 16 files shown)


jagthedrummer:
@gazayas @kaspth @pascallaliberte @andrewculver @olbrich I'd love to hear what y'all think of this.

I'm on the fence about the current structure. I like being able to run individual bits of a workflow manually for testing purposes, and this allows for that. But it also forces everything to be listed in the "Actions" list. It might be better to have a single "CI/CD Pipeline" workflow that contains all of the bits and just handles staging/production stuff differently based on some conditionals or something.

Am I overthinking it about the visual clutter in the list?

jagthedrummer:
Would also be interested in any thoughts about GitHub Actions for internal testing purposes vs it just being a thing that we ship for developer convenience.

gazayas (commented):
@jagthedrummer
> Would also be interested in any thoughts about GitHub Actions for internal testing purposes vs it just being a thing that we ship for developer convenience.

I'm personally ok with using GitHub Actions for this, and either way I think we should have a clear way of separating tests intended for developers from internal tests even if we weren't per se deliberately putting everything in GitHub Actions (i.e. - move Super Scaffolding system tests to `bullet_train-core`, etc). Speaking of which...

> Are the 🏗️ Super Scaffolding Tests even useful for downstream apps? Maybe that should be a separate workflow that's more of an internal tool?

For downstream applications, the only thing I can think of that would really be affected would be routing, if at all. Most of the Super Scaffolding logic consists of flags within the code like this:

```ruby
# 🚅 super scaffolding will insert new fields above this line.
```

However, for routes, we look for method names in the source code directly and adjust the file according to what we find (a big reason why I wanted to use Prism nodes for finding things). I don't think developers should run into any issues since we have `standardrb` in place, but it's just one area where it might cause unforeseen test failures. I'm definitely okay with making the test exclusively internal though, I don't think the routes is _too_ big of an issue to be worried about.

> Do we need to ship these disabled or anything? If someone started a new app and pushed to GitHub would it be bad for a CI pipeline to kick off? (My concern here is about cost / use of free minutes.)

At the very least, we could comment out or remove these tasks on `bin/configure`.

## Other internal tests
Just thinking out loud here, but if were start specifying which tests should be exclusively internal, how hard do we want to draw the line? For example, `test/system/animations_test.rb` seems like something we could make internal, but I don't think we should make `test/system/account_test.rb` internal. Should we mark up a curated list?

.github/workflows/_database_schema_check.yml
```diff
@@ -0,0 +1,27 @@
+# This workflow will ensure that when we run `rails db:migrate` that no changes are made to `db/schema.rb`.
+#
+# This workflow is pimarily meant to be called by other workflows, but it can be run manually.
+name: "🔎 ~ Database Schema Check"
```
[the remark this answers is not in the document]
jagthedrummer: Good catch!

jagthedrummer:
@gazayas I moved the Super Scaffolding tests out of the main pipelines and into a new `🚅 _ BT - Internal CI Pipeline`.

As far as other test being "internal" I'm not sure we really need to mess with that. The super scaffolding tests are _sort of_ a different _type_ of test than the normal test suite. The normal test suite kind of serves two purposes. 1) It gives BT developers a way to know that the functionality they ship is working, and 2) it gives downstream developers a way to know that their app is continuing to work the way the BT developers intended (until they modify the test suite).

The super scaffolding tests really _only_ do that first thing.

I _do_ think there's an argument to be made that the super scaffolding tests really shouldn't be in the starter repo, but should be a part of `core`. Though I'm not sure the extent of starter-repo/core entanglement for super scaffolding. If there are ways that we could break super scaffolding by making a change in the starter repo then those tests might be in the right spot.

gazayas:
> The normal test suite kind of serves two purposes. 1) It gives BT developers a way to know that the functionality they ship is working, and 2) it gives downstream developers a way to know that their app is continuing to work the way the BT developers intended (until they modify the test suite).
The super scaffolding tests really only do that first thing.

I agree. Also about the super scaffolding tests, I'm also currently unsure to what extent we want to shift around the logic between repos. Will keep it in the back of my mind for now.

Just took a look at `🚅 _ BT - Internal CI Pipeline`, cool stuff :+1:

.github/workflows/_reusable_workflows.yml
```diff
@@ -0,0 +1,12 @@
...
+name: " 🚫 |----- Reusable Worfklows Below"
+on:
+  workflow_call:
+
+jobs:
+  nothing:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
kaspth, line 12: Is this sorta needlessly costly to run? I can't remember the syntax but iirc there's some output helpers on GitHub Actions, so maybe we could just print "ok" instead?

.github/workflows/ci-cd-pipeline-main.yml
```diff
@@ -0,0 +1,36 @@
...
+# This workflow is primarily meant to be triggered automatically, but it can be run manually.
+name: " 🎥 Main CI/CD Pipeline (prod)"
+on:
+  workflow_dispatch:
+  push:
+    branches: [ "main" ]
+
+jobs:
+  mini_test:
+    name: 🧪 MiniTest
```
kaspth, line 16: Minor nit but technically the name is [minitest](https://github.com/minitest/minitest/), don't know if we want to lowercase the m though 🙂

```suggestion
  minitest:
    name: 🧪 Minitest
```

kaspth:
> I'm personally ok with using GitHub Actions for this, and either way I think we should have a clear way of separating tests intended for developers from internal tests even if we weren't per se deliberately putting everything in GitHub Actions

Do we have access to the repo and/or organization name in GitHub Actions scripts? I wonder if we could use that to differentiate, so when we're running on `bullet-train-co/bullet_train` we also need to run the internal tests?

> Am I overthinking it about the visual clutter in the list?

I like the way the names look in the sidebar given what we have available to work with. I think it's ok to showcase some of the warts here, since we can't do much else. I'm not sure how the scripts would look if we hid the complexity more within them.

> Would also be interested in any thoughts about GitHub Actions for internal testing purposes vs it just being a thing that we ship for developer convenience.

I mean I'm fine with using it for our internal tests too versus Circle CI since I've found Circle difficult to work with.

.github/workflows/_reusable_workflows.yml
jagthedrummer: I don't really expect anyone to ever actually run this on since it's just there to be a visible divider, but it's probably not a bad idea to make it do as little as possible, just in case.

test/system/super_scaffolding_test.rb
```diff
@@ -284,8 +284,8 @@ def setup
   end
 
   test "OpenAPI V3 document is still valid" do
-    visit "http://localhost:3001/api/v1/openapi.yaml"
-    puts(output = `yarn exec redocly lint http://localhost:3001/api/v1/openapi.yaml 1> /dev/stdout 2> /dev/stdout; rm openapi.yaml`)
+    visit "http://127.0.0.1:3001/api/v1/openapi.yaml"
+    puts(output = `yarn exec redocly lint http://127.0.0.1:3001/api/v1/openapi.yaml 1> /dev/stdout 2> /dev/stdout; rm openapi.yaml`)
```
jagthedrummer, line 288: Apparently `localhost` doesn't resolve inside of GitHub Actions?!?

jagthedrummer:
> Do we have access to the repo and/or organization name in GitHub Actions scripts? I wonder if we could use that to differentiate, so when we're running on `bullet-train-co/bullet_train` we also need to run the internal tests?

Good idea on that, @kaspth! I added a conditional to the super scaffolding test job in the internal pipeline so that it will only run if it's happening inside the starter repo itself.

jagthedrummer:
I also investigated using a conditional to control which environment to deploy to, and that works, but... the UI doesn't immediately show the result of the conditional which gives a wrong impression about what's going to happen.

At first it looks like after the tests pass that it's going to deploy to staging AND production, which might be a little unsettling.

CleanShot 2023-11-16 at 15 43 28

Once the tests pass it kinda gives you an indication that it's not going to deploy to both, but the indicators don't seem to agree.

CleanShot 2023-11-16 at 15 47 33

I like the _idea_ of combining the workflows into a single pipeline, but I kinda think that the "double deploy" UI situation is a deal breaker in terms of being something that we want to ship for use by downstream devs. So I'm currently inclined to leave it as two separate workflows, merge this as it stand currently, and see what happens.

What do y'all think @gazayas & @kaspth?

.github/workflows/_run_super_scaffolding_tests.yml
```diff
@@ -0,0 +1,106 @@
...
+
+jobs:
+  test:
+    name: "🏗️"
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # For super scaffolding tests we need to have exactly 7 runners.
+        ci_runners: [TestSite, Project, 'Project::Step', Insight, 'Personality::Disposition', 'Personality::Observation', TestFile]
```
gazayas, line 17: What are we basing this off of again? (I forgot)

I know we define a list of constants in the system test file itself... Should we have an explicit explanation in the system test that points to this so we know to change it if necessary?

gazayas:
> I like the idea of combining the workflows into a single pipeline, but I kinda think that the "double deploy" UI situation is a deal breaker in terms of being something that we want to ship for use by downstream devs.

Agreed, I think it would cause unneeded stress. I'm not too savvy with GitHub Actions, but the main :movie_camera: pipeline isn't showing up for me? Maybe it's a setting on the repository I'm not familiar with yet.

image

Also, we probably can't do anything about this but I wish we could expand the names here somehow...

image

.github/workflows/_run_super_scaffolding_tests.yml
jagthedrummer: We need exactly 7 runners because we look for node indexes 0-6 in this file:

https://github.com/bullet-train-co/bullet_train/blob/main/test/bin/setup-super-scaffolding-system-test

I'd like to figure out a better way of separating those tests, but probably not as part of this PR.

jagthedrummer:
> the main 🎥 pipeline isn't showing up for me?

@gazayas it's not showing up because that workflow file isn't in `main` and hasn't been run on the repo yet. The staging pipeline is showing up as a result of it having been run on this PR. Once we merge this all of the pipelines that I showed in the screenshots will appear.

jagthedrummer:
> Also, we probably can't do anything about this but I wish we could expand the names here somehow...

Yeah, we don't have any way to give the names more room, but we could potentially shorten some things.

For instance `PR CI/CD Pipeline (staging)` could be shortened in a few different ways.

`PR Pipeline`
`PR / staging`
`PR CI/CD`
etc...

jagthedrummer:
I shortened some of the workflow names. We can always adjust later if we want. I'm gonna go ahead and merge this and we'll see how it goes. 🚀

gazayas:
Awesome!

kaspth:
> I like the idea of combining the workflows into a single pipeline, but I kinda think that the "double deploy" UI situation is a deal breaker in terms of being something that we want to ship for use by downstream devs. So I'm currently inclined to leave it as two separate workflows, merge this as it stand currently, and see what happens.

@jagthedrummer I forgot to get back to you on this, but I agree with where you ended up! Thank you for working on this!

Issue #1082, opened 2023-10-13
Troubles with `device_test`


jagthedrummer:
It seems that `device_test` has a few undesirable side effects.

Now when a system test fails the output isn't very useful.

To demonstrate, I changed this line:

https://github.com/bullet-train-co/bullet_train/blob/398489ec0e2bddca2a875c11f6639f58121cf61e/test/system/webhooks_system_test.rb#L36

To say:

```ruby
click_on "A Button That Doesn't Exist"
```

And then when I run that test this is the output:

```
Error:
WebhooksSystemTest#test_team_member_registers_for_webhooks_and_then_receives_them_on_a_macbook_pro_15_inch:
Capybara::ElementNotFound: Unable to find link or button "A Button That Doesn't Exist"
    test/system/webhooks_system_test.rb:36:in `block in <class:WebhooksSystemTest>'
    test/application_system_test_case.rb:118:in `instance_eval'
    test/application_system_test_case.rb:118:in `block (2 levels) in device_test'


rails test test/application_system_test_case.rb:116
```

The error message itself is accurate and points to the right place. `Unable to find link or button "A Button That Doesn't Exist"` is the message we'd expect to see, and the first line in the stack trace points directly to `test/system/webhooks_system_test.rb:36` which is good.

But the command that it suggests to re-run the test is incorrect. When I run that, it doesn't run a single test.

```
$ rails test test/application_system_test_case.rb:116
🌱 Generating global seeds.
🌱 Generating test environment seeds.
Run options: --seed 58265

# Running:



Finished in 0.000885s, 0.0000 runs/s, 0.0000 assertions/s.
0 runs, 0 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /Users/jgreen/projects/bullet-train-co/bullet_train/coverage. 13 / 234 LOC (5.56%) covered.
```

The line that it's pointing to is inside the `device_test` helper method itself.

https://github.com/bullet-train-co/bullet_train/blob/398489ec0e2bddca2a875c11f6639f58121cf61e/test/application_system_test_case.rb#L114-L121

So then I tried to run just that one single test by using the file and line number reported in the stack trace, but that also doesn't run a test.

```
$ rails test test/system/webhooks_system_test.rb:36
🌱 Generating global seeds.
🌱 Generating test environment seeds.
Run options: --seed 64382

# Running:



Finished in 0.001553s, 0.0000 runs/s, 0.0000 assertions/s.
[...]

PR #1113, opened 2023-10-24
Redirect `device_test` definitions to correct test file.

kaspth:
Fixes https://github.com/bullet-train-co/bullet_train/issues/1082

All this is because Rails' `test` method uses `define_method`, which will report this file as the `source_location`
and not the target test file where the test is actually defined.

`define_method` cannot be passed the correct location, we must use `class_eval` with a source string of Ruby for that.
But using a string means we can't refer to the block, like `define_method(name, &block)` would.

So we inject a module that we define methods via `test` to pass the block to, then we override that
via `class_eval` to point it to the correct location.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -112,17 +112,27 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   #     p display_details
   #   end
   def self.device_test(name, &block)
+    # All this is because Rails' `test` method uses `define_method`, which will report this file as the `source_location`
+    # and not the target test file where the test is actually defined.
+    #
+    # `define_method` cannot be passed the correct location, we must use `class_eval` with a source string of Ruby for that.
+    # But using a string means we can't refer to the block, like `define_method(name, &block)` would.
+    #
+    # So we inject a module that we define methods via `test` to pass the block to, then we override that
+    # via `class_eval` to point it to the correct location.
+    location = caller_locations(1, 1).first
+
+    name = name.remove(/[^\w ]/) # `def` is stricter than `define_method` so we have to scrub characters.
+    @device_test_methods ||= Module.new.tap { include _1 }
+
     @@test_devices.each_key do |device_name|
-      test "#{name} on a #{device_name}" do
-        resize_display # TODO: Figure out why the browser window opens if this is done in `setup`.
-        instance_eval(&block)
-      end
+      test_name = ActiveSupport::Testing::Declarative.instance_method(:test).bind_call(@device_test_methods, "#{name} on a #{device_name}", &block)
+
+      class_eval <<~RUBY, location.path, location.lineno
+        def #{test_name}; resize_display; super; end # Use single line to match source line numbers properly.
+      RUBY
     end
   end
-
-  # The `instance_eval` line in `device_test` adds 2 frames in the backtrace which test errors will erroneously point back to.
-  # So we silence those 2 frames.
-  Rails.backtrace_cleaner.add_silencer { /test\/application_system_test_case\.rb.*?(?:instance_eval|device_test)/.match? _1 }
 
   def resize_for(display_details)
     ActiveSupport::Deprecation.warn <<~END_OF_MESSAGE
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -120,6 +120,10 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     end
   end
 
+  # The `instance_eval` line in `device_test` adds 2 frames in the backtrace which test errors will erroneously point back to.
+  # So we silence those 2 frames.
+  Rails.backtrace_cleaner.add_silencer { /test\/application_system_test_case\.rb.*?(?:instance_eval|device_test)/.match? _1 }
+
   def resize_for(display_details)
     ActiveSupport::Deprecation.warn <<~END_OF_MESSAGE
       `resize_for` is deprecated.
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -128,8 +128,9 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
     @@test_devices.each_key do |device_name|
       test_name = ActiveSupport::Testing::Declarative.instance_method(:test).bind_call(@device_test_methods, "#{name} on a #{device_name}", &block)
 
+      # Standard recommends __FILE__ and __LINE__ + 1 here, which points to `application_system_test_case.rb`. It's wrong.
       # TODO: Figure out why the browser window opens if `resize_display` is done in `setup`.
-      class_eval <<~RUBY, location.path, location.lineno
+      class_eval <<~RUBY, location.path, location.lineno # standard:disable Style/EvalWithLocation
         def #{test_name}; resize_display; super; end # Use single line to match source line numbers properly.
       RUBY
     end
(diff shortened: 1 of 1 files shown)


jagthedrummer:
@kaspth, this only partially fixes #1082. It still leaves the problem where you can't run a single `device_test` because the test runner doesn't seem to see them as tests. I'm on your branch locally and when I try to pass a line number to a test it does this:

```
$ rails test test/system/account_test.rb:22
🌱 Generating global seeds.
🌱 Generating test environment seeds.
Started with run options --seed 16832

Finished in 0.00158s
0 tests, 0 assertions, 0 failures, 0 errors, 0 skips
```

I wonder if we need to find a way to use an actual `test` statement in the test files themselves. Maybe we could pass an option to the `test` method to trigger the device tests?

Maybe instead of:

```ruby
  device_test "user can edit their account" do
    #...
  end
```

Something like this?

```ruby
  test "user can edit their account", devices: true do
    #...
  end
```

(Just hypothesizing here, I really don't know if that would work.)

gazayas:
I sat down with this one for a bit today and was able to delegate the device tests to `test` as @jagthedrummer proposed:

```ruby
# test/application_system_test_case.rb

def self.test(name, devices: false, &block)
  if devices
    @@test_devices.each_key do |device_name|
      name = "#{name} on a #{device_name}"
      resize_proc = Proc.new do
        resize_display
        instance_eval(&block)
      end
      super(name, &resize_proc)
    end
  else
    super(name, &block)
  end
end

# test/system/webhooks_system_test.rb
test "team member registers for webhooks and then receives them", devices: true do
  # ...
end
```

This comes from [Testing::Declarative](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/testing/declarative.rb) module in `ActiveSupport` where the `test` method defines the file's individual tests to run.

The only problem is, it's still causing the `0 tests, 0 assertions` issue, I'm assuming because the line filtering is trying to pick up the tests from the `application_system_test_case.rb` file itself:

```
> rails test test/system/webhooks_system_test.rb:24
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2. (called from <top (required)> at /home/gazayas/work/bt/bullet_train/config/environment.rb:5)
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2. (called from <top (required)> at /home/gazayas/work/bt/bullet_train/config/environment.rb:5)
🌱 Generating global seeds.
🌱 Generating test environment seeds.
Started with run options --seed 22175

Finished in 0.00790s
0 tests, 0 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /home/gazayas/work/bt/bullet_train/coverage. 29 / 288 LOC (10.07%) covered.
```

Seems like a pretty sticky issue. On the surface it seems like we can only get line filtering to work if the tests definitions are only explicitly written in the target tests files we're working with.

The rails [test parser](https://github.com/rails/rails/blob/main/railties/lib/rails/test_unit/test_parser.rb) has the following comments and I figured we could potentially edit this to pick up on `device_test` too, but I'm not sure how far we'll want to dig into rails internals to solve this issue:

```ruby
# Parse a test file to extract the line ranges of all tests in both
# method-style (def test_foo) and declarative-style (test "foo" do)
```

jagthedrummer:
Thanks for looking into this, @gazayas. I wonder if it would be worth opening an issue on `rails` about this.

kaspth:
Ok, I see what's going on here, it's because this `source_location`  https://github.com/rails/rails/blob/1e4f16108401da9952bddf9333b99181f57fe631/railties/lib/rails/test_unit/runner.rb#L172 points to `application_system_test_case.rb` due to how `test` works via `define_method`. I'm trying out a potential fix, but it's kinda tricky.

kaspth:
Nope, that's not it entirely, and there's something going on with the line filtering too.

kaspth:
Ok, I just pushed a fix. It's a little tricky due to how Ruby works.

But with this I can get `rails test test/system/webhooks_system_test.rb:36` from #1082 to work again. And test failures point to the correct test file, not `application_system_test_case.rb`.

I've tried to explain what's going on in a comment, since it's fairly in-depth unfortunately.

jagthedrummer:
Holy smokes, nice work on that @kaspth!

kaspth:
@jagthedrummer thank you, I went ahead and merged then!

PS: we would've had the same issue if we swapped to `test "name", devices: true` since that would also call `super` from within `application_system_test_case.rb` and thus `define_method` would point there. Although, I technically didn't verify that.

gazayas:
@kaspth Awesome stuff, just confirmed it's working for me!

gazayas:
> PS: we would've had the same issue if we swapped to test "name", devices: true since that would also call super from within application_system_test_case.rb and thus define_method would point there. Although, I technically didn't verify that.

Yes, I confirmed that earlier by prying into [definition_for](https://github.com/rails/rails/blob/1e4f16108401da9952bddf9333b99181f57fe631/railties/lib/rails/test_unit/runner.rb#L180-L182) which previously returned the line number from the application system test case file:

```
> git diff test/system/webhooks_system_test.rb
diff --git a/test/system/webhooks_system_test.rb b/test/system/webhooks_system_test.rb
index 445a4fcf..5ac2c98f 100644
--- a/test/system/webhooks_system_test.rb
+++ b/test/system/webhooks_system_test.rb
@@ -21,7 +21,7 @@ class WebhooksSystemTest < ApplicationSystemTestCase
     Rails.application.reload_routes!
   end
 
-  device_test "team member registers for webhooks and then receives them" do
+  test "team member registers for webhooks and then receives them", devices: true do
     login_as(@user, scope: :user)
 
     visit root_path
> bundle exec rails test test/system/webhooks_system_test.rb:24
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2. (called from <top (required)> at /home/gazayas/work/bt/bullet_train/config/environment.rb:5)
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2. (called from <top (required)> at /home/gazayas/work/bt/bullet_train/config/environment.rb:5)
🌱 Generating global seeds.
🌱 Generating test environment seeds.
Started with run options --seed 58831

/home/gazayas/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/binding_of_caller-1.0.0/lib/binding_of_caller/mri.rb:21: warning: constant RubyVM::DebugInspector is deprecated

Frame number: 0/10

From: /home/gazayas/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/railties-7.1.2/lib/rails/test_unit/runner.rb:183 Rails::TestUnit::Filter#definition_for:

    182: def definition_for(method)
 => 183:   binding.pry
    184:   TestParser.definition_for(method)
    185: end

[1] pry(#<Rails::TestUnit::Filter>)> method
=> #<UnboundMethod: WebhooksSystemTest#test_team_member_registers_for_webhooks_and_then_receives_them_on_a_macbook_pro_15_inch() /home/gazayas/work/bt/bullet_train/test/application_system_test_case.rb:136>
```

kaspth:
@gazayas just to clarify: the problem isn't so much with `definition_for`, it's the `method_obj.source_location` within that was pointing to the wrong file/line. So the trickery we did here was to fix that root cause.

kaspth:
You can see that where you're printing `method` above, `definition_for` hasn't touched it and it's already pointing to `application_system_test_case.rb`.

PR #1108, opened 2023-10-21
Implement home page

gazayas:
Closes [#615.](https://github.com/bullet-train-co/bullet_train-core/issues/615)

To view the page:
1. Comment out `include RootRedirect` in the [home controller](https://github.com/bullet-train-co/bullet_train/blob/main/app/controllers/public/home_controller.rb).
2. Add an `index` action in the same controller.

Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/628

https://github.com/bullet-train-co/bullet_train/assets/10546292/3bb3240a-0627-42c8-be12-899a8eb7ff05

PR #1143, opened 2023-11-10
Use home page by default

gazayas:
Continuation of #1108.

There were no issues with tests thankfully due to #990 and #1007.
Only one problem where I had to make sure the `Signed out successfully` flash notice showed up properly.

### DIFF
--- a/app/controllers/public/home_controller.rb
+++ b/app/controllers/public/home_controller.rb
@@ -1,11 +1,14 @@
 class Public::HomeController < Public::ApplicationController
-  # Redirect `/` to either `ENV["MARKETING_SITE_URL"]` or the sign-in page.
-  # If you'd like to customize the action for `/`, you can remove this and define `def index ... end ` below.
-  include RootRedirect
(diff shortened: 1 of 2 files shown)


jagthedrummer:
I'm thinking we should do a larger version bump on this one since it'll change the behavior of apps if people don't notice the change and undo it for themselves.

jagthedrummer:
Or... maybe it should be a config option that lives as turned off in the starter repo but that we turn on via `bin/setup`? I thought we were doing that for some config option or another but I'm not finding it right off the bat.

gazayas:
> I thought we were doing that for some config option or another but I'm not finding it right off the bat.

Yeah, I don't THINK we have a config in place, but I'm impartial as to having one or not. I know there are developers who have their [marketing site somewhere else](https://discord.com/channels/836637622432170028/1100107180033716394/1100178346198962237), so I can whip up an implementation in `bin/configure` if that's ideal. Otherwise it's currently not clear that developers can edit this setting manually when setting up a new app.

PR #1151, opened 2023-11-13
Remove Knapsack Pro

### LINKED ISSUE #1124: Potentially remove knapsack pro
There were some changes to CircleCI and parallelism somewhat recently, and we might not have a need to run things with knapsack pro any longer. Also, the pricing has changed as well, so it might not be worth the cost for our current setup.

gazayas:
Closes #1124.

We decreased our parallelism recently [here](https://github.com/bullet-train-co/bullet_train/pull/976), and with the current state of things we won't be needing Knapsack Pro anymore.

I believe there's an environment variable on CircleCI, so if this gets cleared we'll have to delete that as well. I'm also currently working on a PR to remove all knapsack pro remains from `bullet_train-core` (I don't think those tests are actually being run, but either way we should remove the Knapsack Pro lines that are still there).

The tests are currently failing due to an OpenAPI issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,21 +118,11 @@ jobs:
       - *wait_for_docker
 
       - run:
-          name: Run tests with Knapsack Pro
-          command: |
-            export RAILS_ENV=test
-            bundle exec rails "knapsack_pro:queue:minitest[--verbose]"
-          environment:
-            KNAPSACK_PRO_CI_NODE_TOTAL: 2
-
-      # If you don't want to use Knapsack Pro, then use this configuration:
-      #
-      # - run:
-      #     name: Run unit tests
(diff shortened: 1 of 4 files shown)


gazayas:
Joint PR
- https://github.com/bullet-train-co/bullet_train-core/pull/657

jagthedrummer:
I think we need to get a 👍 / 👎 from @andrewculver on this one. I believe that ClickFunnels uses Knapsack Pro internally with higher parallelism than we use for testing the starter repo, so they may not want us removing it from the CircleCI config. Maybe if we move to GitHub actions for BT internal stuff we just don't add Knapsack Pro there?

gazayas:
Haven't heard about this one in a while, so I'll go ahead and close it. We can handle it if necessary later.

PR #1155, opened 2023-11-15
Work on invitation lists tests again

gazayas:
#1028.

Just going to leave this one open as a draft for now, but I'm still thinking it's related to how Capybara responds to `assert_text` after redirects. This time I've added a couple of `assert_text`s beforehand instead.

I've rerun multiple times (seven times to be exact) on the last commit, but again, will just leave this open as a draft for now and re-experiment later.

By the way, here is some debugging information which looks at the page text. The first image is from the first commit in this PR, and the second image from the second commit.

Capybara 2

Capybara 1

### DIFF
--- a/test/system/invitation_lists_test.rb
+++ b/test/system/invitation_lists_test.rb
@@ -5,6 +5,7 @@ class InvitationListsTest < ApplicationSystemTestCase
     # TODO: Capybara.current_driver is returning :selenium,
     # when it should be either :selenium_chrome or :selenium_chrome_headless.
     Capybara.current_driver = Capybara.default_driver
+    p Capybara.current_driver
 
     @current_bulk_invitations_setting = nil
     BulletTrain.configure do |config|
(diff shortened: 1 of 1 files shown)


gazayas:
After more experimentation, besides the invitation lists test itself, there was an error in `invitations_test.rb` (oversight on my part) → In the flow of this test, we didn't need to check for the `Skip` button because bulk invitations are only available for the first membership on the teams, so I deleted the line because the button doesn't show up in the first place.

https://github.com/bullet-train-co/bullet_train-core/blob/3383ce9c8e461b2b682f8c321ae41db900084797/bullet_train/app/controllers/concerns/account/onboarding/user_details/controller_base.rb#L30

gazayas:
Even if this doesn't fix #1028, we can at least merge it for `invitations_test.rb`.

jagthedrummer (approved):

PR #1202, opened 2023-11-27
Write more precise db schema check

gazayas:
Again this is failing due to our current CI credit status, but the db schema check works for me. I wrote a comment in https://github.com/bullet-train-co/bullet_train/pull/1150 explaining a little bit in more detail.


gazayas:
Just to make sure it's clear, this is pointing to `depfu/update/group/rails-7.1.2` and not `main`.

.circleci/db_schema_check
```diff
@@ -10,6 +10,34 @@ echo $GIT_STATUS
...
+
+  # Reset the settings we edited above.
+  unset IFS
+  set +f
+
+  if [[ ${#ADDITIONS[@]} == 1 && ${ADDITIONS[0]} =~ ^\+ActiveRecord ]] ; then
+    echo "Rails version was updated, but there were no changes to any tables."
+    exit 0
+  fi
+
```
kaspth, line 40: I don't remember how to use an exit status after a command, but I think we could simplify this to something like the following:

```suggestion
  `git diff db/schema.rb | grep "^+[^+]" | grep -v "ActiveRecord::Schema\["`

  if [[ $? == 1 ]] ; then
    echo "Rails version was updated, but there were no changes to any tables."
    exit 0
  fi
```

Basically we leverage that addition lines start with `+` and then we filter out `ActiveRecord::Schema[`, `grep` will return 1 if there's no matches — meaning that no other changes were made in the file.

What happens if all that changes in a schema file is the `version:`? That's of course pretty rare, but do we end up failing the build in that case?

gazayas: Hey @kaspth, thank you for the review! This is definitely a lot simpler. I didn't know about `$?` returning a status, that's pretty useful. 

I tested this to check if a developer were to update their rails version AND and a model which they forget to migrate in the same go (an unlikely situation but possible nonetheless), and the check here covers that, so I'll try to push this change here soon.

kaspth: > I tested this to check if a developer were to update their rails version AND and a model which they forget to migrate in the same go (an unlikely situation but possible nonetheless), and the check here covers that, so I'll try to push this change here soon.

Ok, sounds good!

kaspth (approved):
Nice!

PR #1206, opened 2023-11-28
[CI] Run CORE tests in GitHub Actions

jagthedrummer:
This makes it so that we run tests against the `core` repo as part of our internal pipeline.

CleanShot 2023-11-28 at 12 27 17

(CircleCI is giving us trouble (again!), so it might be time to migrate away.)

### DIFF
--- a/.github/workflows/_run_super_scaffolding_tests.yml
+++ b/.github/workflows/_run_super_scaffolding_tests.yml
@@ -91,13 +91,7 @@ jobs:
           CIRCLE_NODE_INDEX: ${{ strategy.job-index }}
 
       - name: 'Run Super Scaffolding Test'
-        run: bundle exec rails test test/system/super_scaffolding_test.rb
-
-      - name: 'Run Super Scaffolding Partial Test'
-        run: bundle exec rails test test/system/super_scaffolding_partial_test.rb
-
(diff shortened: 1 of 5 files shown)


.github/workflows/_run_super_scaffolding_tests.yml
```diff
@@ -48,23 +58,34 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          path: tmp/starter
+
+      - name: Checkout Core Repo
```
gazayas, line 64: I see that we have `Link Core Repo` in `_run_tests.yml`, should we have that here too? Is this file fine without it?

jagthedrummer: Yep, we do need that here as well! Thanks for the review and keen eye, @gazayas.

jagthedrummer:
Since tests were auto-failing in CircleCI due to billing issues I looked for a way to just disable Circle without completely ripping everything out yet. To do that I just disabled the webhook that GitHub sends to Circle when something happens.

In the repo settings I went to the Webhooks area and edited the CircleCI webhook:

CleanShot 2023-11-29 at 09 52 20

And then unchecked the "Active" checkbox:

CleanShot 2023-11-29 at 09 52 27

Issue #1174, opened 2023-11-21
Different handling for `HIDE_THINGS`


jagthedrummer:
Currently we have CI configured to do two entire runs of the test suite, one with `HIDE_THINGS=true` and one with `HIDE_THINGS=false`. And then a bunch of tests either run, or don't depending on that ENV var.

It would be nice if we could do just a single run of the test suite, and have things setup so that `HIDE_THINGS` can be toggled on or off as needed for specific tests.

Part of the challenge is that we currently end up evaluating `HIDE_THINGS` in `config/routes.rb`, which is generally only parsed once, at the beginning of a test run. Maybe instead of trying to have route disappear based on that setting we should continue to allow those paths to be routeable, but have the controller reject them somehow?

olbrich:
@jagthedrummer There is a `reload_routes!` method on `Rails::Application` that may help with this.

PR #1207, opened 2023-11-28
[CI] Handle HIDE_THINGS differently in tests

jagthedrummer:
Fixes: https://github.com/bullet-train-co/bullet_train/issues/1174

Joint PR: https://github.com/bullet-train-co/bullet_train-core/pull/701

This PR makes it so that we don't have to run the test suite twice with both possible values of `HIDE_THINGS`. Instead of skipping tests if `HIDE_THINGS` is set to `true`, this makes it so that we set `HIDE_THINGS` to `false` before running those tests, and then we restore `HIDE_THINGS` to it's previous value.

The only functional potentially user-facing change is in the joint PR.

### DIFF
--- a/test/concerns/fields/super_select_support_concern_test.rb
+++ b/test/concerns/fields/super_select_support_concern_test.rb
@@ -1,67 +1,74 @@
 require "test_helper"
 
-unless scaffolding_things_disabled?
-  class Fields::SuperSelectSupportConcernTest < ActiveSupport::TestCase
-    class FakeController < Account::ApplicationController
-      include Fields::SuperSelectSupport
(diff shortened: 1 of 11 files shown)


kaspth (approved):
Cool, this is probably a lot simpler too!

Do we need to write an update script like we did for `device_test` so that end users don't end up with a potentially difficult diff when upgrading to a new BT version?

kaspth:
Ok, good points, let's skip the update script then :shipit:

gazayas (changes requested):
Besides the `AccountTest` issue I left a comment about, everything looks great to me!

PR #1255, opened 2023-12-28
support for docker

VPetar:


VPetar:
Added this as an accident, wanted to merge this into my fork
However, decided still to leave it as a closed PR, since maybe someone would wanna write this a bit better than me :)

PR #1257, opened 2024-01-01
Add comment in GitHub Actions concerning `RAILS_MASTER_KEY`

### LINKED ISSUE #1168: GitHub actions should export the RAILS_MASTER_KEY into the environment
https://github.com/bullet-train-co/bullet_train/blob/0211217230642a8f8f960391bcdbe700e9179c00/.github/workflows/_database_schema_check.yml#L27-L31

If using the rails credentials system, it is necessary to export the RAILS_MASTER_KEY into the environment of runners so that they can decrypt the encrypted credential files.

gazayas:
Closes #1168.

I didn't add this to every environment, such as the PostgreSQL database setup in the following, but I can go ahead and add comments there too if necessary:

https://github.com/bullet-train-co/bullet_train/blob/8925e85a56f33df151df9623b120ef03a87d0994/.github/workflows/_database_schema_check.yml#L13-L20

@cc/ @olbrich

### DIFF
--- a/.github/workflows/_database_schema_check.yml
+++ b/.github/workflows/_database_schema_check.yml
@@ -26,6 +26,8 @@ jobs:
           --health-retries 5
     env:
       RAILS_ENV: test
+      # Manually export your local RAILS_MASTER_KEY if using the credentials system.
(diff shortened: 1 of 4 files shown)


jagthedrummer (approved):
Thanks for handling this, @gazayas!

PR #1263, opened 2024-01-05
Ignore `rails-erd` output

gazayas:
rails-erd generates a PDF when creating and migrating a new database, so I added the file to `.gitignore`.

Also as I was looking over `.gitignore`, I saw that we remove `/storage` completely:

https://github.com/bullet-train-co/bullet_train/blob/911b1f8d1ac92f4ca81f5dbfe0d41e264fe30310/.gitignore#L56-L57

I thought about deleting the following lines altogether, but developers may want to revert to Rails' standard handling of these files, so I commented out these lines and added an explanation.

https://github.com/bullet-train-co/bullet_train/blob/911b1f8d1ac92f4ca81f5dbfe0d41e264fe30310/.gitignore#L19-L24


.gitignore
```diff
@@ -16,12 +16,16 @@
...
-/storage/*
-!/storage/.keep
-/tmp/storage/*
-!/tmp/storage/
-!/tmp/storage/.keep
+# /storage/*
+# !/storage/.keep
+# /tmp/storage/*
+# !/tmp/storage/
+# !/tmp/storage/.keep
```
jagthedrummer, line 28: I'm not sure we want to do this. I think by default we _should_ ignore things in `/storage` because that's not a dependable place for things to be stored and if people aren't paying attention they could end up committing large test files to their repo which can then be a pain to undo later. @gazayas can you talk me through your thought process here? Why are you thinking that people would want to have `/storage` in the repo?

I agree that we should ignore `erd.pdf`.

jagthedrummer (changes requested):
Left one comment/question inline about removing the `/storage` bits.

.gitignore
gazayas: Yeah, the reason I commented these out is because we just ignore the <i>entire</i> `/storage` folder later on in the file in our Bullet Train-specific list.

https://github.com/bullet-train-co/bullet_train/blob/911b1f8d1ac92f4ca81f5dbfe0d41e264fe30310/.gitignore#L56-L57

I figured we shouldn't delete the Rails standard ignore lines though just in case anyone wants to default to them, so I tried to let developers know that they can just delete `/storage` from our list at the bottom and uncomment these lines if they want to do so. Without the comment they might be confused as to why the whole folder is being ignored if they just look at lines 25-28 on the surface and miss out on line 57.

That's just my thought process, it's not a hill I'm willing to die on 😃 Will revert the changes if we just want to ignore erd instead.

jagthedrummer: Ah-ha! I had missed the duplication. I think if I were making the call that I'd remove the duplicate from lines 56/57 and leave the rails defaults active. Theirs are more specific (they ensure that the storage dir won't ever go missing during a repo clone because of being empty) and their lines cover `/tmp/storage` as well as just `/storage`.

gazayas: Sounds good, just updated the file.

jagthedrummer (approved):

PR #1247, opened 2023-12-19
Allow local css overrides

pascallaliberte:
Resolves #567.

Joint PR:
* https://github.com/bullet-train-co/bullet_train-core/pull/726

PR #1274, opened 2024-01-12
Add default local stylesheet

### LINKED ISSUE #567: Add way to have local css for overrides
Currently, the main tailwind-powered css file in found inside the current theme's gem. It's parsed by tailwind and included into the `<head>`, all via the theme's own files.

https://github.com/bullet-train-co/bullet_train/blob/17d557a5434d143f9b7cc77536bc9516b8a35a0c/package.json#L37

* There should be a way to have a batteries-included local css file for custom overrides. 
* These overrides should be loaded after all Tailwind generated classes
* The local css file should allow for npm-powered `@import` statements (already permitted via `postcss.config.js` `postcss-import`.
* Ideally, this should allow the use of `@apply` directives that know about classes defined in all the files parsed via Tailwind's `content[]` config.

pascallaliberte:
Resolves #567.
Replaces #1247.

Introduces a new local stylesheet:

`app/assets/stylesheets/application.css` loads `$ThemeStylesheet`, an alias that postcss-import resolves to the current theme's stylesheet, responsible for initializing tailwind. It also loads...

Both `postcss.config.js` and `postcss.mailer.config.js` are updated with a `postcssImportConfig`, which resolves the new `$ThemeStylesheet` alias. This allows us to continue not requiring to publish an npm package for each theme.

Joint PR:
* https://github.com/bullet-train-co/bullet_train-core/pull/741

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1,0 +1,13 @@
+/* The theme's stylesheet initializes tailwindcss and imports the theme's default styles.
+
+   The $ThemeStylesheet path is evaluated via a postcss resolver, which finds the theme's
+   main stylesheet within the theme's gem.
+   
+   To control the order of the tailwindcss imports further, you can eject the theme's
+   stylesheet locally using the following command, which will take precedence.
+   > rake bullet_train:themes:light:eject_css
+
+   Read more: https://bullettrain.co/docs/stylesheets
+*/
+@import "$ThemeStylesheet";
+@import "application.custom";
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -10,4 +10,8 @@
    Read more: https://bullettrain.co/docs/stylesheets
 */
 @import "$ThemeStylesheet";
-@import "application.custom";
+
+/* ✅ YOUR APPLICATION'S CSS */
+/* If you need to customize your application's CSS, this is the place to do it. This helps
+   avoid merge conflicts in the future when Rails or Bullet Train update their own CSS.
+*/
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "scripts": {
     "build": "THEME=\"light\" node esbuild.config.js",
     "build:css": "bin/link; THEME=\"light\" yarn build; yarn light:build:css; yarn light:build:mailer:css",
-    "light:build:css": "THEME=\"light\" NODE_PATH=./node_modules tailwindcss -c tailwind.config.js -i $(bundle exec bin/theme tailwind-stylesheet light) -o ./app/assets/builds/application.light.css --postcss ./postcss.config.js",
+    "light:build:css": "THEME=\"light\" NODE_PATH=./node_modules tailwindcss -c tailwind.config.js -i ./app/assets/stylesheets/application.css -o ./app/assets/builds/application.light.css --postcss ./postcss.config.js",
     "light:build:mailer:css": "NODE_PATH=./node_modules tailwindcss -c $(bundle exec bin/theme tailwind-mailer-config light) -i $(bundle exec bin/theme tailwind-stylesheet light) -o ./app/assets/builds/application.mailer.light.css --postcss ./postcss.mailer.config.js"
   }
 }
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "scripts": {
     "build": "THEME=\"light\" node esbuild.config.js",
-    "build:css": "THEME=\"light\" yarn build; yarn light:build:css; yarn light:build:mailer:css",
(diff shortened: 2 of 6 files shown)


pascallaliberte:
To do:

* [ ] make sure ejected themes (with no published gems) will still resolve the `$ThemeStylesheet` alias (see https://github.com/bullet-train-co/bullet_train/pull/1247#pullrequestreview-1801291641)

gazayas:
@pascallaliberte I took a quick look at things and was able to add my own custom styling to `application.custom.css` without a problem. Looks great!

I ran into these issues when trying to a eject a new theme, it looks like the `postcss-import-config.js` file isn't resolving properly:

```
foo-css        | bundler: failed to load command: bin/theme (bin/theme)
foo-css        | bin/theme:54:in `<top (required)>': Sorry, we couldn't find `/postcss-import-config.js` in any of the following paths: /Users/gabriel_zayas/work/bt/bullet_train (RuntimeError)

...

foo-css        | Error: Command failed: bundle exec bin/theme postcss-import-config foo
foo-css        | bundler: failed to load command: bin/theme (bin/theme)
foo-css        | bin/theme:54:in `<top (required)>': Sorry, we couldn't find `/postcss-import-config.js` in any of the following paths: /Users/gabriel_zayas/work/bt/bullet_train (RuntimeError)

**Comment by gazayas** [comment]
Ok yes, I'm seeing how it's related, I wrote a potential solution here https://github.com/bullet-train-co/bullet_train-core/pull/741/files#r1452382879

File: app/assets/stylesheets/application.css [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+   main stylesheet within the theme's gem.
+   
+   To control the order of the tailwindcss imports further, you can eject the theme's
+   stylesheet locally using the following command, which will take precedence.
+   > rake bullet_train:themes:light:eject_css
+
+   Read more: https://bullettrain.co/docs/stylesheets
+*/
+@import "$ThemeStylesheet";
+@import "application.custom";
```
kaspth, line 13: Why do we need the separate `application.custom` file?

I'm thinking of the pattern we've established in the rest of the starter repo like this:

```ruby
class User < ApplicationRecord
  include Users::Base
end
```

My meaning being that in the starter repo, users own the resulting files but we integrate by pulling stuff in from the rest of Bullet Train. Having a separate `application.custom.css` breaks that (albeit subtle) convention. Versus `application.css` having a theme import while also containing custom extensions makes more sense to me. What do you think?

pascallaliberte: Yeah, I was having second thoughts about `application.custom` for the same reasons as we're breaking the pattern. (And I thought you'd be bringing it up @kaspth).

Even [`app/javascript/application.js`](https://github.com/bullet-train-co/bullet_train/blob/main/app/javascript/application.js) is where we encourage people to edit after our own imports.

https://github.com/bullet-train-co/bullet_train/blob/ebc520f1bb56d990aa263b88be69ff08429283d4/app/javascript/application.js#L22-L34

Will remove the second file.

kaspth: @pascallaliberte oh cool! I didn't even think about `app/javascript/application.js`, but looks like there's a pattern we can follow in there with similar wording.

> (And I thought you'd be bringing it up @kaspth).

You know me too well 😄

pascallaliberte:
@jagthedrummer I'll wrap it up next week

pascallaliberte:
@jagthedrummer the `postcss-import-config.js` errors are likely (correct me if I'm wrong) caused by tests that don't match up the branch names between both repos, could it be?

package.json
```diff
@@ -37,7 +37,7 @@
   "scripts": {
     "build": "THEME=\"light\" node esbuild.config.js",
     "build:css": "bin/link; THEME=\"light\" yarn build; yarn light:build:css; yarn light:build:mailer:css",
```
jagthedrummer, line 39:
```suggestion
    "build:css": "THEME=\"light\" yarn build; yarn light:build:css; yarn light:build:mailer:css",
```

I'd like to see what happens in CI if we remove this call to `bin/link`. I think it's clobbering the branch-matched core gems that we link earlier in the CI process. (Also, I don't understand why we'd ever want that call to `bin/link` there in the first place. It seems like an internal-BT-dev tool thing that we shouldn't need as part of normal operation.)

kaspth: `bin/link` is part of normal operation for Bullet Train apps, it's how CSS/JS assets get resolved from engines/gems. Since the gems get symlinked into `tmp/gems`, the Tailwind config can then pick them up: https://github.com/bullet-train-co/bullet_train-core/blob/dab52ad22c932a4c7003c8747ba1c668c39f2444/bullet_train-themes-light/tailwind.light.config.js#L11-L14

I'm not sure what the alternatives for engine assets would be, if we want to avoid a too large prebuilt Tailwind compiled CSS file.

jagthedrummer (approved):

PR #1275, opened 2024-01-16
Bundle capybara-lockstep to improve System Test stability.

kaspth:
https://github.com/makandra/capybara-lockstep

Seems like it costs 5 seconds, but the build fails in strange ways. So I'm opening this for posterity.

I also used `bundle open bullet_train-themes-light` to add the expected JS include in `app/views/themes/light/layouts/_head.html.erb`:

```diff
+<%= capybara_lockstep if defined?(Capybara::Lockstep) %>
<%= javascript_include_tag 'application', 'data-turbo-track': 'reload' %>
```

@gazayas @jagthedrummer I'll probably close this unless either of you feel like diving into the test failures, because I don't know what they mean 😊

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,10 @@ group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara", "~> 3.39"
 
+  # Synchronize Capybara commands with client-side JavaScript and AJAX requests to greatly improve
+  # system test stability. Only works on the Selenium Driver though.
+  # gem "capybara-lockstep"
+
   # Selenium is the default default Capybara driver for system tests that ships with
   # Rails. Cuprite is an alternative driver that uses Chrome's native DevTools protocol
   # and offers improved speed and reliability, but only works with Chrome. If you want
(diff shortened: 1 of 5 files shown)


jagthedrummer:
Oh, this looks interesting. I didn't know about `capybara-lockstep`. I actually think the test failure on this PR is a timing issue that might be addressed by `capybara-lockstep` if it weren't commented out of the `Gemfile`.

gazayas:
I like the idea of including this because I've run into many `visit`-related test failures in the past, and it would be nice not to have to think about them. I'll try to keep this in mind so I can look into the test failures soon.

jagthedrummer:
@kaspth I un-commented the gem in the `Gemfile` to activate it, and added the required JS snippet to `app/views/layouts/_head.html.erb` and verified that the integration is working by briefly un-commenting the debug line in `test/application_system_test_case.rb`. It seems to be doing the thing. I've re-run tests on this PR a number of times and none of the occasionally-flaky tests flaked out, so... maybe this does the trick? 🤞

I think it's worth merging and giving it a shot. Is there anything else you think we should do on this?

kaspth:
> I think it's worth merging and giving it a shot. Is there anything else you think we should do on this?

@jagthedrummer sounds good, let's give it a shot 😄

Issue #1284, opened 2024-01-18
Need better defaults for ActiveStorage to allow for production deployment


jagthedrummer:
If you start a new BT project and immediately push it to Heroku (or if you update a previously existing BT project to rails 7.1 _AND_ you haven't configured an S3 bucket for ActiveStoratge) the app will fail to boot with this error:

```
Missing Active Storage service name. Specify Active Storage service name for config.active_storage.service in config/environments/production.rb
```

That's due to the interaction of a change [introduced to rail in this PR](https://github.com/rails/rails/pull/44372) and this code in the BT starter repo:

https://github.com/bullet-train-co/bullet_train/blob/cd59fffa796574589a47650b23b34a1c48d4f1c9/config/environments/production.rb#L112-L114

If you haven't set up ENV vars with S3 keys then we don't configure any storage service, and then the change from the PR linked above kicks in and throws an error before the app even boots.

I think we need to either just configure the `:amazon` service even if the keys are missing (unless that also throws and immediate error) or we should configure the `:local` service. If we end up configuring the `:local` service we should probably also output a bit ol' banner letting people know that we fell back to the local service but that it's a really bad idea to use it in production unless you know that you have persistent storage connected to your web server.

PR #1289, opened 2024-01-19
Allow the app to boot if S3 is not configured

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train/issues/1284

There may be better ways to handle this, but right now anyone who updates to a new version (or who starts a new project) is going to have trouble even booting their app in a production environment if they haven't configured S3.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -111,6 +111,9 @@ Rails.application.configure do
 
   if (ENV["AWS_ACCESS_KEY_ID"] || ENV["BUCKETEER_AWS_ACCESS_KEY_ID"]).present?
     config.active_storage.service = :amazon
+  else
+    puts "WARNING! : We didn't find an active_storage.service configured so we're falling back to the local store, but it's A VERY BAD IDEA to rely on it in product unless you know what your'e doing."
+    config.active_storage.service = :local
   end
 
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -111,6 +111,9 @@
 
   if (ENV["AWS_ACCESS_KEY_ID"] || ENV["BUCKETEER_AWS_ACCESS_KEY_ID"]).present?
     config.active_storage.service = :amazon
+  else
+    puts "WARNING! : We didn't find an active_storage.service configured so we're falling back to the local store, but it's A VERY BAD IDEA to rely on it in product unless you know what your'e doing."
```
zackgilbert, line 115: Nit: think you mean: "production"?

jagthedrummer: I did, indeed! Thanks for the sharp eye.

zackgilbert (approved):
This is great. Thank you! 

Like I mentioned in Discord, a nice additional would be a link to a BT specific doc helping people easily set this up (or directing them to a service that'll do this for them).

jagthedrummer:
Created a separate issue for making some docs that we can link to. Gonna go ahead and merge this just to try to minimize the number of people who stub their toe on this.

PR #1300, opened 2024-01-24
Visit Creative Concepts page directly in reactivity system test

### LINKED ISSUE #1288: Reactivity system test could be more specific
This test assumes that the creative concepts list is visible on the default root path. When this isn't the case the test fails. I think we could have the test visit the creative concepts index page first, and then do everything from that screen and we'd be testing what we want to test without making assumptions about what lives on the root path.

test/system/reactivity_system_test.rb

gazayas:
Closes #1288.

### DIFF
--- a/test/system/reactivity_system_test.rb
+++ b/test/system/reactivity_system_test.rb
@@ -36,9 +36,13 @@ class ReactivitySystemTest < ApplicationSystemTestCase
       end
     end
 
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
Thanks, @gazayas!

gazayas:
No problem!

Issue #833, opened 2023-08-09
Refactor super scaffolding tests


jagthedrummer:
In this conversation (https://github.com/bullet-train-co/bullet_train-core/pull/392#discussion_r1289155629) @kaspath says:

I wonder if we should refactor these tests into a `test/system/super_scaffolding` directory on the starter repo end? So we could just get `bundle exec rails test test/system/super_scaffolding/` here.

Could do `test/system/super_scaffolding/super_scaffolding_test.rb` and `test/system/super_scaffolding/partial_test.rb` for the test names.

gazayas:
Sounds good to me too!

I was also thinking of editing the setup script so we super scaffold only select models depending on the flags we pass to `bundle exec test/bin/setup-super-scaffolding-system-test`. Sometimes it can be a hassle to run the entire setup script when I just went to test one model/functionality locally. We can add that later if anything though.

PR #1311, opened 2024-01-30
Move scaffolding system tests

gazayas:
Closes https://github.com/bullet-train-co/bullet_train/issues/833

This doesn't create flags for isolating specific models to generate in the setup script like I mentioned in the comment [here](https://github.com/bullet-train-co/bullet_train/issues/833#issuecomment-1672282384), but it's not too high-priority so I'm okay with closing the issue out with just this PR.


gazayas:
We could also just change `super_scaffolding_test.rb` to `base_test.rb`

PR #1316, opened 2024-02-02
Remove `device_test` fix for logged in users

gazayas:
`device_test` has undergone some changes, and although this was raising an error previously when running the entire test suite, no errors are being raised now.

### DIFF
--- a/test/system/fields_test.rb
+++ b/test/system/fields_test.rb
@@ -15,13 +15,6 @@ class FieldsTest < ApplicationSystemTestCase
   end
 
   device_test "simulate restoring behavior of form fields on page restore" do
-    # TODO: This is an ugly hack to ensure the default device_test user isn't logged in.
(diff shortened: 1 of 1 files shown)


gazayas:
I think it's also worth noting that if running the test suite as a whole is causing users to stay logged in between tests, we would see the same behavior in tests like these which access the root path before going to the registration page:

https://github.com/bullet-train-co/bullet_train/blob/37099726978eac62388a66e242a40314cf14f09e/test/system/authentication_test.rb#L4-L8

https://github.com/bullet-train-co/bullet_train/blob/37099726978eac62388a66e242a40314cf14f09e/test/system/membership_test.rb#L4-L8

Checking here right after `visit root_path` and checking the pages though, users aren't logged in to begin with, so I think this bug is gone for now.

jagthedrummer (approved):
My favorite kind of PR! Deleting code.

PR #1329, opened 2024-02-12
Run super scaffolding tests via directory

kaspth:
Follow-up to #1311

With the new directory we can ensure any new super scaffolding test file is automatically run, so we don't forget.


jagthedrummer (approved):
Nice.

PR #1348, opened 2024-03-02
Pull latest changes for depfu bundler update

gazayas:
Rebase isn't yielding the results we want in #1313, so I did `merge main` here to make the tests pass.


jagthedrummer:
Hmmm... this seems weird. I wonder why rebasing isn't doing the right thing. I'm a little hesitant to merge this just because of the fact that depfu will automatically merge that other PR once tests start passing, even if there's other weirdness.

Issue #1440, opened 2024-05-06
`bin/configure` could be broken up


jagthedrummer:
The `bin/configure` script does a lot of things and it's not really possible (or at least not easy) to run only a portion of the script. It might be nice to extract the various pieces out into separate scripts that could be run individually.

The biggest problem with the current arrangement is that we first do a lot of work to try to detect and install dependencies, before moving on to making changes to the app config and setting up git. We don't gracefully handle dependency detection on all systems, and when that fails it can end up blocking people from being able to finish making the config changes needed.

So, in addition to extracting the bits into separate scripts we might also revisit the order in which we run those individual scripts. We might want to do config tweaking first, and then move on to dependency installation.

Related to https://github.com/bullet-train-co/bullet_train/issues/1435

jagthedrummer:
Starting to take a look at what all happens in `bin/configure` and am just going to kind of live-tweet my findings here.

The first thing we do is make sure that a couple of utility gems are installed and required.

https://github.com/bullet-train-co/bullet_train/blob/f4a32e769990e2700b4f1d6fe29d2b2655e40416/bin/configure#L3-L12

Then we define several utility functions that we use later in the script.

https://github.com/bullet-train-co/bullet_train/blob/f4a32e769990e2700b4f1d6fe29d2b2655e40416/bin/configure#L14-L61

Then we check that the version of ruby on the local system matches what we expect. If we don't have the right version we don't just bail out automatically, we give the user the option to try to continue with whatever ruby version they're using.

https://github.com/bullet-train-co/bullet_train/blob/f4a32e769990e2700b4f1d6fe29d2b2655e40416/bin/configure#L65-L78

Next we check whether `homebrew` is installed. We do this check even on Windows and Linux, which is unnecessary because those platforms don't have `homebrew`. This should be moved to be a part of the MacOS only conditional branch below.

https://github.com/bullet-train-co/bullet_train/blob/f4a32e769990e2700b4f1d6fe29d2b2655e40416/bin/configure#L80-L88

Now we start a `case` statement where we do various things depending on your local OS.

https://github.com/bullet-train-co/bullet_train/blob/f4a32e769990e2700b4f1d6fe29d2b2655e40416/bin/configure#L90

[...]

Issue #1435, opened 2024-04-27
`bin/configure` assumes Debian based system


coder2000:
I use Arch Linux so my package manager is `pacman` or `yay` if I need to use something from AUR. The configure script as provided by the repo assumes that `dpkg` is available wants to install `postgresql` and `redis` system wide. While that isn't an issue in and of itself I prefer to use docker to run them on a per project basis if needed.

coder2000:
Splitting it up didn't resolve the issue. The issue is that the scripts rely primarily on `dpkg` and `apt-get` which are only available on Debian and derivatives.

jagthedrummer:
@coder2000 Can you please post a stack trace or some details about whatever problem you ran into? The new scripts no longer rely on `dpkg` or `apt-get` for detecting postgres or redis. It should attempt to use `dpkg` for detecting `icu4c` but if your system doesn't have `dpkg` it _should_ just tell you that it knows you don't have `dpkg` and then it should suggest that you ensure that `icu4c` is installed.

https://github.com/bullet-train-co/bullet_train/blob/e2f4afac4ebd780b7332a134567386a2bcccfbd1/bin/configure-scripts/check_icu.rb#L78-L81

What exactly are you seeing?

coder2000:
Ok. Sorry, I missed that part of the script.

jagthedrummer:
@coder2000 if it gives you any problems please report back with some details. We definitely want to make this as smooth as possible.

BRGustavoRibeiro:
Hi! Same problem here on Fedora 40 Sway. 

- **dpkg** is installed through `brew`.
- `libicu` installed on system.
- **icu4c** installed through `brew`.

I honestly have no idea how to fix this.

`You don't have dpkg, so we can't tell if icu4c is installed.`

pushcx:
Here’s the last couple dozen lines of `bin/setup` crashing on Arch:

```
--------------------------------------------------------------------------------
Checking PostgreSQL

You have PostgreSQL installed, but you're using v16.3 and not v14.
Try proceeding without PostgreSQL 14? [Y/n]

Continuing with PostgreSQL v16.3. This might cause problems in the next steps.


--------------------------------------------------------------------------------
Checking Redis


You have redis 7.2.5 installed.


--------------------------------------------------------------------------------
Checking icu4c


[...]

Issue #1412, opened 2024-04-01
`bin/configure` should probably rename `origin` even if you skip GitHub setup


jagthedrummer:
If you do this:

```
git clone git@github.com:bullet-train-co/bullet_train.git your_new_project_name
cd your_new_project_name
bin/configure
```

It will ask you if you want to push your project to GitHub. If you say no, it will leave the starter repo as `origin` in your local repo.

We probably don't ever want a downstream project to consider the starter repo as `origin`. We probably want to rename it `upstream` regardless of how you answer the GitHub question.

Issue #1388, opened 2024-03-29
Repo link in README is not replaced


newstler:
Some why [this code](https://github.com/bullet-train-co/bullet_train/blob/c73c9c843f73bb935279b844d59619d9a6b4b560/bin/configure#L272) doesn't change the repo in README.

jagthedrummer:
I think this is related to the fact that we had some bad commits on `main`.

jagthedrummer:
I just looked into this and I think it's a legit bug.

There's this line where we ask if you want to setup GitHub:

https://github.com/bullet-train-co/bullet_train/blob/c73c9c843f73bb935279b844d59619d9a6b4b560/bin/configure#L188

And then on the next line we test it like this:

https://github.com/bullet-train-co/bullet_train/blob/c73c9c843f73bb935279b844d59619d9a6b4b560/bin/configure#L189

But then further down the script we do some less specific tests, that don't really test what we want to be testing.

https://github.com/bullet-train-co/bullet_train/blob/c73c9c843f73bb935279b844d59619d9a6b4b560/bin/configure#L269

https://github.com/bullet-train-co/bullet_train/blob/c73c9c843f73bb935279b844d59619d9a6b4b560/bin/configure#L283

jagthedrummer:
Relatedly, it seems that we only even attempt to replace the repo URL if you **did not** choose to setup GitHub. Which seems less than ideal.

https://github.com/bullet-train-co/bullet_train/blob/c73c9c843f73bb935279b844d59619d9a6b4b560/bin/configure#L269-L273

If you _do_ setup GitHub we ask you for the `ssh` variant of the repo URL and not the `https` variant. But [according to the render docs](https://docs.render.com/deploy-to-render#specifying-a-repository) they need the `https` variant.

> Set the value of repo to the full https:// URL for your Git repo.

I think maybe we should explicitly ask if you want to use Render. If you're planning to deploy to Heroku or literally any other deployment target it doesn't really do you any good to have a "deploy to Render" button on your README. In fact, it could really be counterproductive and send people down a dead end.

PR #1478, opened 2024-05-21
Split up `bin/configure` into a series of smaller scripts

jagthedrummer:
This is a first pass at splitting `bin/configure` into a series of smaller scripts. It also tries to fix a few issues that we have open against `bin/configure`.

To test this out you can clone this particular branch from the starter repo and then run the `bin/configure-beta` script. If you wanted to name the test project `bin_configure_beta_test` you could do this:

```
git clone --single-branch -b jeremy/bin-configure-beta git@github.com:bullet-train-co/bullet_train.git bin_configure_beta_test
cd bin_configure_beta_test
./bin/configure-beta
```

For now I've made `configure-beta` a separate script so that you could also run `bin/configure` after checking out the repo for easy comparison. Before merging this PR we probably want to update it to _replace_ `bin/configure` with the new version.

Fixes: https://github.com/bullet-train-co/bullet_train/issues/1440
Fixes: https://github.com/bullet-train-co/bullet_train/issues/1435
Fixes: https://github.com/bullet-train-co/bullet_train/issues/1412
Fixes: https://github.com/bullet-train-co/bullet_train/issues/1388
Fixes: https://github.com/bullet-train-co/bullet_train/issues/814


bin/configure-beta
```diff
@@ -0,0 +1,48 @@
...
+# First we require a util file that will install a few utility gems
+# and define some utility methods used by other scirpts.
+require "#{__dir__}/configure-scripts/utils"
+
+# Then we check that we have the right ruby version
+require "#{__dir__}/configure-scripts/check_ruby"
+
+# We rename `origin` to `bullet-train` for consistency with documented upgrade
+# procedures. Also to prevent any unintended pushes to the starter repo.
+require "#{__dir__}/configure-scripts/rename_origin"
```
gazayas, line 12: :point_up: :100:

bin/configure-scripts/push_to_origin.rb
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
+
+require "#{__dir__}/utils"
+
+has_origin_remote = `git remote | grep origin`.strip.length > 0
```
gazayas, line 5: @jagthedrummer

I changed `origin` to `bullet-train` remote at the beginning as the script suggested, and near the end of the script I ran into this error:

```
Would you like to commit the changes to your project? [Y/n]
n

Make sure you save your changes with Git.
This repo doesn't appear to have an `origin` remote configured. Skipping pushing to `origin`.
/home/gazayas/work/bt/test_apps/bin_configure_beta_test/bin/configure-scripts/check_postgres.rb:5:in ``': No such file or directory - postgres (Errno::ENOENT)
	from /home/gazayas/work/bt/test_apps/bin_configure_beta_test/bin/configure-scripts/check_postgres.rb:5:in `<top (required)>'
	from <internal:/home/gazayas/.rbenv/versions/3.3.1/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:136:in `require'
	from <internal:/home/gazayas/.rbenv/versions/3.3.1/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:136:in `require'
	from ./bin/configure-beta:32:in `<main>'
> 
```

I see that you've added this line in another file, so maybe we should add it here too?
```ruby
has_bt_remote = `git remote | grep bullet-train`.strip.length > 0
```

gazayas (commented):
I made one inline comment, but besides that this looks much more manageable! Thanks for working on this!

bin/configure-scripts/push_to_origin.rb
jagthedrummer: Thanks for giving it a shot @gazayas! I don't think this error actually had anything to do with the missing `origin`. It's related to some refactoring I did in trying to detect postgres.

It's complaining about this line:

https://github.com/bullet-train-co/bullet_train/blob/904c7bc0c087082be9bd0f3296665f5ebd858529/bin/configure-scripts/check_postgres.rb#L5

When I'm just trying to run the `postgres --version` command to detect if it's installed and what version. I've now wrapped that in a `begin`/`rescue` block so that it won't error out if the `postgres` command isn't able to be found. And I added a fallback to check for `psql` if `postgres` isn't found. (I also added some "section headers" to make it more clear which step of the config process we're on at any given time.) Can you give it another shot?

Also, I'm curious what platform are you running on? I had hoped we could get away from having to maintain platform and package-manager specific ways of detecting postgres (and redis & icu4c), but maybe that's wishful thinking...

bin/configure-scripts/utils.rb
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env ruby
+
+puts ""
+puts `gem install colorize`
+puts ""
+
+require "colorize"
+require "active_support"
```
kaspth, line 8: Do these indicate we actually want something like this?

```ruby
require "bundler/inline"

gemfile do
  gem "colorize"
  gem "activesupport", require: "active_support"
end
```

bin/configure-beta
kaspth, line 1: All this stuff is really interesting!

Though I'm wondering if we should just have a Ruby class to embed all the scripts in, sorta like this?

```ruby
# bin/configure.rb
class Configure
  def check_ruby
    announce_section "Checking Ruby"

    # Unless the shell's current version of Ruby is the same as what the application requires, we should flag it.
    # rbenv produces strings like "3.1.2" while rvm produces ones like "ruby-3.1.2", so we account for that here.
    required_ruby = `cat ./.ruby-version`.strip.delete_prefix("ruby-")
    actual_ruby = `ruby -v`.strip

    message = "Bullet Train requires Ruby #{required_ruby} and `ruby -v` returns #{actual_ruby}."
    colorize = actual_ruby.include?(required_ruby) ? :green : :red
    puts message.then(&colorize)
    check_proceed! "Try proceeding with Ruby #{actual_ruby} anyway?"
  end

  private

  def announce_section(name)
    # …
  end

  def check_proceed!(question)
    input = ask "#{question} [y/n]"
    exit 1 unless input.downcase.first == "y"
  end
end
```

(Also includes a slight refactoring of `check_ruby`)


And then we'd use it like this:

```ruby
# bin/configure
#!/usr/bin/env ruby

require_relative "configure"

conf = Configure.new
conf.check_ruby
conf.more_things
```

kaspth: Another option is to use a Rails generator for this, maybe with something like:

```ruby
require "bundler/inline"

gemfile do
  gem "railties"
end

require "rails/generators"

class ConfigureGenerator < Rails::Generators::Base
  # Since generators are built on Thor::Groups, they'll run every public method in sequence.
  #
  # We also get access to all the standard Thor::Actions
  # https://www.rubydoc.info/gems/thor/Thor/Actions
 
  # …as well as Rails' Generators::Actions.
  # https://api.rubyonrails.org/classes/Rails/Generators/Actions.html
  #
  # Some of our utils/helpers reimplement things we'd get from generators.
end
```

garrettdimon: I don't know that I'd re-write all of this, but quickly scanning the code, it seems to be recreating a lot of what generators provide. It doesn't seem to need templates anywhere, so that makes make it relatively simple to use generators for this.

But in particular, `insert_into_file`, `append_to_file`, `prepend_to_file`, `gem`, `run`, `ask`, `yes?`, and `no?` etc. would likely do a lot of the heavy lifting and enable deleting a good amount of code. I've got a page that pulls together the Thor and [Rails actions into a single page](https://garrettdimon.com/journal/posts/rails-generator-actions#gsub_file) for browsing as it might help show some other actions that would be useful.

kaspth: @garrettdimon thank you!

@jagthedrummer I'm not saying we need to consider switching to a generator in the same go as splitting things out, or even at all. I don't want to put work on your plate so if you'd rather continue with the script file based approach, please go ahead 😊

bin/configure-scripts/utils.rb
jagthedrummer: Yeah, I think that's a great suggestion. Thanks!

bin/configure-beta
jagthedrummer: Thanks for the suggestion, @kaspth. And thanks for giving it a look @garrettdimon.

I think for this pass I don't want to bite off too much at once. My main goals in splitting this up are:
1. Making it possible to run (or re-run) just a portion of the config script without having to jump through a bunch of hoops or manually disable code.
2. Making it easier to grok what exactly is involved for each particular "step".

I don't think that putting everything in to a single `Configure` class is going to work for those goals.

But generators might be a great way to encapsulate some of this, and the built in helper functions that Garrett mentions would definitely be useful in reducing the amount of code that we own and have to maintain. Since you can call other generators and script from within a generator I think we could potentially replace some of theses "sub scripts" with generators piece by piece in the future.

bin/configure
```diff
@@ -1,295 +1,48 @@
 #!/usr/bin/env ruby
 
-puts ""
```
jagthedrummer, line 3: The entire diff for this file looks really messy because I extracted various parts of it into sub scripts and changed the order in which things happen. So some of the parts of the diff look like they don't make any sense because it'll look like I removed a block about checking the ruby version and replaced it with a sub script for checking postgres.

PR #1481, opened 2024-05-23
Update gitpod.Dockerfile to ruby version 3.3.1

### LINKED ISSUE #1427: Gitpod starts with ruby 3.2.2, but bulletrain requires 3.3.0
Steps to reproduce:
1. Click on gitpod icon on repo homepage
2. Select default gitpod settings
3. See errors in terminal

PeterMonsson:
This fix closes #1427. Previous problems seen in #1427 seem to have been fixed on the gitpod side of things

### DIFF
--- a/gitpod.Dockerfile
+++ b/gitpod.Dockerfile
@@ -1,7 +1,7 @@
 FROM gitpod/workspace-postgres
 USER gitpod
 
-RUN _ruby_version=ruby-3.2.2 \
+RUN _ruby_version=ruby-3.3.0 \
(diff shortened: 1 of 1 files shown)


jagthedrummer (approved):
Thanks @PeterMonsson!

jagthedrummer:
@PeterMonsson, for the yarn/corepack problem I _think_ the solution is to just run `corepack enable` prior to running `yarn install`. The `gitpod.Dockerfile` seems to abstract all of that away so I'm not sure where exactly that should be done. (Also, I'm not a gitpod user.) If you could try it out and let us know if you could get it to work that would be great.

PR #1514, opened 2024-06-05
Fix method name in `check_redis.rb`

citizen428:
This PR fixes an incorrect method name in `bin/configure-scripts/check_redis.rb` from `ask_boolan` to `ask_boolean`.


jagthedrummer (approved):
Thanks, @citizen428!

Issue #1524, opened 2024-06-07
We shoud activate jemalloc by default on heroku


jagthedrummer:
We can add the appropriate buildpack to `app.json` and set the needed ENV var.

Issue #1523, opened 2024-06-07
A fresh app instantly runs out of memory on heroku


jagthedrummer:
We should probably set `WEB_CONCURRENCY` and `RAILS_MAX_THREADS` to be lower.

We could potentially change this directly in `config/puma.rb`.

Or it might be better to set them as environment variables in `app.json`.

Issue #1521, opened 2024-06-07
Including Sentry in `app.json` breaks review apps on heroku


jagthedrummer:
If you try to use review apps with the current `app.json` Heroku raises an error when trying to create the review app that says:

```
Sentry has opted out of supporting Review and CI apps
```

Related to: https://github.com/bullet-train-co/bullet_train/issues/1127

Issue #1127, opened 2023-10-31
Can we slim down the price tag associated with `app.json`


jagthedrummer:
On the [docs page for deploying](https://bullettrain.co/docs/heroku) we say this:

> When you're ready to deploy to Heroku, it's highly recommended you use this button:
>
> [Deploy to Heroku]
>
> This button leverages the configuration found in app.json, including sensible production-ready defaults for dyno formation, third-party services, buildpack configuration, etc. The resources provisioned will cost about $140/month.

Personally I would strongly recommend _against_ people using that button because I think it _wildly_ over-provisions an app for what _most_ people want when first deploying an app.

I'm not sure we should go for "production-ready". I'd go for something more like "show off the MVP" as the starting point.

My recommendations are:

* Set the dyno types to `basic` instead of `standard-1x`. Default dyno cost would go from $100/mo to $14/mo.
* Set the postgresql add-on to `mini` instead of `standard-0`. Default DB cost would go from $50/mo to $5/mo.
* Set the redis addon to `mini` instead of `premium-0`. Default Redis cost would go from $15/mo to $3/mo.
* Set the memcachier addon to `dev` instead of `100. Default memcache cost would go from $15/mo to $0/mo.
* Remove `bucketeer` completely. Unless I'm missing something this seems like a fairly irresponsible add-on for us to suggest. It seems like just an inflated price tag for S3. The `hobbyist` plan included in `app.json` costs $5/mo, which allows you to store a whopping 5GB of data in S3. The price tag for that much raw S3 storage is about 12 cents per month. At the high end they charge $1,250 for 1TB of storage which should cost $23 when paying AWS directly. If we remove this that's $5 off the default cost. (And we'd not be setting people down a questionable road.)

All together this would take the default Heroku deployment cost from ~$185/mo (which is more than the $140 that we currently have documented) to a much more reasonable ~$22/mo.

Relatedly, do we need both Honeybadger and Sentry? Seems like maybe we should pick just one.

PR #1525, opened 2024-06-07
Better & cheaper initial experience on Heroku

jagthedrummer:
This PR does a few things to make the process of setting up an app on Heroku better and cheaper.

1. It adds a buildpack for `jemalloc` and sets the `JEMALLOC_ENABLED` ENV var. This helps with memory usage. Fixes: https://github.com/bullet-train-co/bullet_train/issues/1524
2. Makes the initial setup cheaper by using cheaper dynos and the entry level for a number of addons (in addition to removing the Bucketeer addon that seems questionably useful to begin with). This brings the monthly cost (as provisioned by `app.json`) down from ~$185/month to ~$22/month. Fixes: https://github.com/bullet-train-co/bullet_train/issues/1127
3. Removes the Sentry addon. This was redundant since Honeybadger was already included, and Sentry isn't supported in review apps on Heroku, so having it in `app.json` prevents the use of review apps entirely which is frustrating. Fixes: https://github.com/bullet-train-co/bullet_train/issues/1521
4. Sets `RAILS_MAX_THREADS` and `WEB_CONCURRENCY` to match the values [suggested by heroku](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#recommended-default-puma-process-and-thread-configuration). Fixes: https://github.com/bullet-train-co/bullet_train/issues/1523

Joint PR for udpating docs: https://github.com/bullet-train-co/bullet_train-core/pull/850

### DIFF
--- a/app.json
+++ b/app.json
@@ -6,28 +6,27 @@
   "keywords": ["ruby", "rails"],
   "buildpacks": [
     { "url": "heroku/nodejs" },
+    { "url": "https://github.com/gaffneyc/heroku-buildpack-jemalloc.git" },
     { "url": "heroku/ruby" }
   ],
   "formation": {
     "web": {
       "quantity": 1,
-      "size": "standard-1x"
+      "size": "basic"
     },
     "worker": {
       "quantity": 1,
-      "size": "standard-1x"
+      "size": "basic"
     }
   },
   "addons": [
-    "heroku-postgresql:standard-0",
-    "heroku-redis:premium-0",
-    "memcachier:100",
-    "bucketeer:hobbyist",
+    "heroku-postgresql:essential-0",
+    "heroku-redis:mini",
+    "memcachier:dev",
     "cloudinary:starter",
     "scheduler:standard",
     "rails-autoscale:white",
-    "honeybadger:basic",
-    "sentry:f1"
+    "honeybadger:basic"
   ],
   "env": {
     "BASE_URL": {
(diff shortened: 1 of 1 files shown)


jagthedrummer:
We could potentially take another $7 off the monthly cost by defaulting to embedded Sidekiq instead of having a `worker` dyno.

app.json
```diff
@@ -46,6 +45,18 @@
       "description": "This ensures the Super Scaffolding templates don't show up in production.",
       "generator": "secret",
       "required": true
+    },
+    "RAILS_MAX_THREADS": {
+      "description": "The maximum number of threads that puma will fork. https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#recommended-default-puma-process-and-thread-configuration",
+      "value": "5"
```
kaspth, line 51: Rails 8 is going to be shifting to a default of 3, I think we should suggest the same despite Heroku's defaults. I can't remember if they did a more condensed write up of it, but there was quite a long thread of discussion on GitHub.

kaspth (approved):

kaspth:
> We could potentially take another $7 off the monthly cost by defaulting to embedded Sidekiq instead of having a `worker` dyno.

This definitely sounds interesting too! I don't know how embedded is to run though since I'm not much of an Ops person.

app.json
jagthedrummer: @kaspth I'd be interested in learning more about that discussion. I also tested this number with the demo site and it's working well and keeping things withing the memory quota. What benefits do you see with defaulting to 3 instead of 5?

jagthedrummer: Gonna go ahead and merge this as-is for now. We can always circle back later and change it if we need to.

PR #1642, opened 2024-08-03
Test latest nice_partials; couldn't get CI working there, so we're do…

kaspth:
…ing this


kaspth:
Eyy, success! We can close this out and release NicePartials now.

PR #1650, opened 2024-08-10
Sketch out a BulletTrain::FormBuilder

kaspth:
Having your templates littered with `render "shared/fields/<form_field>", form:, method:` and all the options managing doesn't feel very Rails-like.

Since Rails already has an API built-in for extending the form builders we could leverage that.

This is by no means complete, I just wanted to open this and show that it's possible for now.


jagthedrummer:
@kaspth I love this idea! I agree that rendering partials directly feels a little un-Railsy.

Issue #1663, opened 2024-08-22
Localization test is flaky (maybe mainly in CI)


jagthedrummer:
In one of my projects I often get errors in `test/controllers/application/localization_test.rb`.

I notice it most often in GitHub Actions.

The errors look like this:

```
ApplicationControllerTest
  test_team_locale_is_nil,_user_locale_is_nil                    ERROR (0.42s)
Minitest::UnexpectedError:         ActionController::UrlGenerationError: No route matches {:action=>"any_action", :controller=>"application"}
            test/controllers/application/localization_test.rb:35:in `block in <class:ApplicationControllerTest>'

  test_user_not_signed_in_and_browser_sends_HTTP_ACCEPT_LANGUAGE_with_unknown_valueERROR (0.11s)
Minitest::UnexpectedError:         ActionController::UrlGenerationError: No route matches {:action=>"any_action", :controller=>"application"}
            test/controllers/application/localization_test.rb:96:in `block in <class:ApplicationControllerTest>'
```

When it happens there are _always_ 9 errors coming from that test file. Often it's the only failure. Just re-running the tests usually doesn't help, but pushing an empty commit does. This may to indicate some sort of ordering problem.

PR #1664, opened 2024-08-22
Fix flaky localization tests

kaspth:
Fixes https://github.com/bullet-train-co/bullet_train/issues/1663

We shouldn't replace the whole Rails' app's RouteSet, particularly when the `@routes` instance variable already provides the hook.

Instead we can mirror what Rails does internally:
https://github.com/rails/rails/blob/a59ddc714fa0403cbd8c7ae6abdec2f71ee7b63f/actionpack/test/controller/test_case_test.rb#L1224C1-L1236C8


jagthedrummer (approved):
Thanks, @kaspth!

PR #1689, opened 2024-09-11
Add script to check for image processing deps

hopsoft:
If `vips` or `imagemagick` is unhealthy, misconfigured, or missing, it can block the completion of `bin/setup`. In the case of an unhealthy or misconfigured system it can cause all Rails scripts to segfault and exit instantly without guideance on how to proceed.

This PR addresses the concern by ensuring the user has these dependencies installed on their system and that they're working before proceeding.

- imagemagick
- vips

```
ruby bin/setup-scripts/check_image_dependencies.rb

--------------------------------------------------------------------------------
Checking Image Processing Dependencies


Checking vips...
vips is installed: 

Checking ImageMagick...
ImageMagick is not installed or not found in PATH.
Try `brew install imagemagick` on macOS or `sudo apt-get install imagemagick` on Ubuntu.
Would you like to continue without ImageMagick? [y/N]

You've chosen not to continue without ImageMagick. Goodbye
```


jagthedrummer:
Thanks for this @hopsoft! There's not really any reason to have both of them installed is there? If not, maybe we should check for both of them first, and if we find one, just proceed? And if both are missing then show a message for how to install one or the other? Or both? 🤔

PR #1708, opened 2024-10-09
Update Gitpod setup

PeterMonsson:
Updates the Gitpod Dockerfile and Gitpod startup scripts so that Gitpod boots seamlessly.


jagthedrummer (approved):
Thanks, @PeterMonsson!

Issue #1609, opened 2024-07-15
"Deploy to Heroku" button doesn't work from created repo


andrewculver:
<img width="264" alt="image" src="https://github.com/user-attachments/assets/d1474eef-1c2c-4020-b98e-404b291efdf6">

jagthedrummer:
Maybe we just remove the "Deploy to Heroku" button for a configured repo? Or somehow check to see if the repo is public before asking if they want the button?

gazayas:
> Maybe we just remove the "Deploy to Heroku" button for a configured repo?

I actually like this idea. The original README serves as a primer for getting started with Bullet Train, so it makes sense to me that developers would look at the README and apply any additional configuration (like these deploy buttons) if it's necessary for their use-case, even after the initial setup. We don't necessarily <i>need</i> to have the buttons exist in a new application, as long as we point developers in the right direction.

It may be beneficial to just have solid documentation on how to deploy with each service and link to it in the README too. That would also help to cut down on things to do in `bin/configure` which would make it quicker/smoother for developers to get setup with a new application.

If we <i>do</i> want the buttons though, I submitted #1614 so developers don't have to write in their repository link.

jagthedrummer:
@andrewculver this should be working now. (Assuming you have your Heroku account connected to your GitHub account.)

PR #1727, opened 2024-10-30
Remove addition of deploy buttons from `bin/configure`

jagthedrummer:
The "deploy to X" buttons rely on the target repo being publicly accessible which seems to be rare. So including these buttons tends to only create confusion and doesn't make anything smoother. Removing them will prevent people from turning down this dead end.

I'll update the "deploy to heroku" docs to contain more info about how to deploy a new app.

Fixes: https://github.com/bullet-train-co/bullet_train/issues/1609


jagthedrummer:
Hmm... as I started working on beefing up the instructions for Heroku I realized that there's really no way to provision all of the things from `app.json` without using a "Deploy to Heroku" button. And I found that you can point at a private repo if you're signed in to heroku and have connected your GitHub account. So maybe we want to bring this back and update some things to make it work better.

PR #1810, opened 2024-12-15
Make turned off redis/pg non-breaking during setup

RichStone:
It's possible that a dev installs Postgres and Redis as prompted which will pass the installed check. However, if the dev has those services still turned off, they will fail later during the install, which can lead to confusion, frustration and less activations on BT.


jagthedrummer:
Awesome, @RichStone!

Once this is merged it should fix https://github.com/bullet-train-co/bullet_train/issues/955 & https://github.com/bullet-train-co/bullet_train/issues/954

RichStone:
> Awesome, @RichStone!
> 
> Once this is merged it should fix #955 & #954

Oh that's cool 💪

PR #1821, opened 2024-12-17
Ability to super scaffold fields onto Membership

jagthedrummer:
This makes it possible to super scaffold new fields onto the `Membership` model.

For instance:

```
rails g super_scaffold:field Membership description:text_field
```

The new field renders on the `Membership` form:

CleanShot 2024-12-17 at 14 21 54

And on the `Membership` `show` page:

CleanShot 2024-12-17 at 14 22 17

It's also added to the `Invitation` form:

CleanShot 2024-12-17 at 14 23 45

And is passed through the Invitation to the Membership:

CleanShot 2024-12-17 at 14 24 19

Joint PR: https://github.com/bullet-train-co/bullet_train-core/pull/990

### DIFF
--- a/app/controllers/account/invitations_controller.rb
+++ b/app/controllers/account/invitations_controller.rb
@@ -1,0 +1,11 @@
+class Account::InvitationsController < Account::ApplicationController
+  include Account::Invitations::ControllerBase
+
+  private
+
+  def permitted_membership_fields
+    [
+      # 🚅 super scaffolding will insert new fields above this line.
+    ]
+  end
+end
--- a/app/controllers/account/invitations_controller.rb
+++ b/app/controllers/account/invitations_controller.rb
@@ -4,8 +4,9 @@ class Account::InvitationsController < Account::ApplicationController
   private
 
   def permitted_membership_fields
-    [
-      # 🚅 super scaffolding will insert new fields above this line.
-    ]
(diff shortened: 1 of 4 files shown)


app/controllers/account/invitations_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  private
+
+  def permitted_membership_fields
+    # We render app/views/account/memberships/_attributes.html.erb on the Invitation form
+    # so by default we permit all of the fields that the MembershipController permits. If
+    # you want more control you can remove the following line and then return an array of
+    # membership attributes that you want to allow at the time of iniviation. For example:
+    # [ :description ]
+    Account::MembershipsController.new.send(:permitted_fields)
```
jagthedrummer, line 12: Potentially we could do this in `core` and either remove this method, or maybe even the entire file. I kinda think it makes sense to expose it here and give people an easier way to override the default.

app/views/account/memberships/_attributes.html.erb
```diff
@@ -0,0 +1 @@
+<%# 🚅 super scaffolding will insert new fields above this line. %>
```
jagthedrummer, line 1: This is rendered by `memberships/show.html.erb` (which is in `core`).

Issue #1774, opened 2024-11-21
Rails  8 Question


BrettFraley:
Perhaps this is not the place to ask, but what do future plans to migrate and upgrade to Rails 8 look like at this point?

Thanks in advance for any feedback .

jagthedrummer:
Hey @BrettFraley we'll definitely be updating to Rails 8 soonish. We've been giving it a little time in the wild to allow it to get "fully baked" before making the switch. I can't promise an exact timeline but it will likely be before the end of the year, or if not, early next year.

BrettFraley:
@jagthedrummer Thank you! That's great news! It gives me enough time to establish what I'm building now and be able to segway into adopting Bullet Train as needed.

PR #1824, opened 2024-12-18
Rails 8.0.1

jagthedrummer:
This updates the Starter Repo from Rails `7.2` to `8.0`. It also activates the new framework defaults in `config/initializers/new_framework_defaults_8_0.rb`. If you have trouble with this update it's probably worth checking that file to see if any of the config options there relate to the problem you're seeing.

Fixes https://github.com/bullet-train-co/bullet_train/issues/1774


BrettFraley:
@jagthedrummer Thanks! I'm considering this a definite xmas gift!

PR #1952, opened 2025-02-19
fix: avo `time_zone` options

### LINKED ISSUE #1371: Avo team & user view don't work in production environment
There is currently an issue with the team and user view of the avo dashboard in production:

image

The issue is, that in these views, a select_field (from avo) is used to display a time_zone value. The [select_field](https://github.com/avo-hq/avo/blob/432ef75df3ca10e79eb953f176ee4d22f8020e48/lib/avo/fields/select_field.rb#L57) uses `options.invert[value]`.

In the [User](https://github.com/bullet-train-co/bullet_train/blob/main/app/avo/resources/user.rb) and [Team](https://github.com/bullet-train-co/bullet_train/blob/main/app/avo/resources/team.rb) model, these `options` get fed by `view_context.time_zone_options_for_select(Avo::Current.user.time_zone, nil, ActiveSupport::TimeZone)`, which returns a `ActiveSupport::SafeBuffer(String)`. This class doesn't implement an `invert` method.

In the development environment this works, because the `colorize` gem gets included, which implements this method for the String class. But in production, this gem/method is missing.

Paul-Bob:
Fixes #1371

This PR fixes the `time_zone` select field options on Avo resources.

Previously, the code:

```ruby
view_context.time_zone_options_for_select(Avo::Current.user.time_zone, nil, ActiveSupport::TimeZone)
```

was generating a `ActiveSupport::SafeBuffer`, which is not compatible with the select field `options` that expects a hash.

### DIFF
--- a/app/avo/resources/team.rb
+++ b/app/avo/resources/team.rb
@@ -15,7 +15,9 @@ class Avo::Resources::Team < Avo::BaseResource
     field :name, as: :text
     field :slug, as: :text
     field :being_destroyed, as: :boolean, hide_on: :forms
-    field :time_zone, as: :select, options: -> { view_context.time_zone_options_for_select(Avo::Current.user.time_zone, nil, ActiveSupport::TimeZone) }
+    field :time_zone, as: :select, options: -> {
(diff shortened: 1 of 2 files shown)


jagthedrummer (approved):
Thank you!

PR #1986, opened 2025-03-11
Fix openai documentation url

zackgilbert:
We were pointing to the wrong gem documentation page (openai-ruby vs ruby-openai). Took me a minute to realize it wasn't correct.


jagthedrummer (approved):
Good catch, thanks!

PR #1994, opened 2025-03-15
Add Kamal to template repo

RichStone:
So I got this to a point where it is working for me enough so that I can do my thing. I'm still up for a few iterations to make this nice for any BT user, but I wanted to get this out already and let it simmer. And also take a short break from Kamal deployments. 😅

There are also _levels_ to how nice, automated, and secure we want the setup to be. In the current state, we lay a "95%-there foundation" for the user to have a quick start, still leaving a few manual changes up to the user: They'll need to fix a few files after bin/configure with their IP address and such. But even if we get those 5% out of the way, the user will still need to do extra work to get the setup and their servers really right, so I'm not sure we need to automate and improve much more (though, _we could_).

There is an accompanying setup article here:

- [richstone.io/rails-8-on-kamal](https://richstone.io/rails-8-on-kamal-example-setup-with-sidekiq-redis-and-postgres).

Not sure where it should go, but I'm okay with moving the source of truth instructions somewhere else for this (BT docs? BT instructions? BT blog?). Or nudge the PR in a way that it doesn't need instructions at all. 🙈


Dockerfile
RichStone, line 1: I'm basically still using the one Dockerfile that I got from swombat in the BT Discord because it's the one I got running but I think it needs slimming down?

I've thought more about something like [this](https://github.com/chap/rails-8-beta1/blob/main/Dockerfile). But didn't spend the time to divide and conquer stuff that's needed and not needed here since it mostly worked from the start.

It also needs updating of ruby/node versions and we probably want to make them dynamic based on the `.ruby-version` file?

bin/configure-scripts/deploy_kamal_instructions.rb
```diff
@@ -0,0 +1,19 @@
...
+if add_instructions
+  File.open("README.md", "a") do |readme|
+    instruction = File.read(instructions_file)
+    instruction.each_line do |line|
+      readme << line
+    end
+  end
+else
+  puts "Not adding 'Deploy to Kamal' instructions.".yellow
+end
```
RichStone, line 19: We could probably add many more questions here, like "do you want to set up Kamal now?" and then prompt for the IP, etc. and auto-populate many more things.

bin/docker-entrypoint
```diff
@@ -0,0 +1,15 @@
...
+
+# Enable jemalloc for reduced memory usage and latency.
+if [ -z "${LD_PRELOAD+x}" ]; then
+    echo "🚀 Saving RAM with jemalloc!"
+    LD_PRELOAD=$(find /usr/lib -name libjemalloc.so.2 -print -quit)
+    export LD_PRELOAD
+fi
+
+echo "Running ./bin/rails db:prepare in docker-entrypoint..."
+./bin/rails db:prepare
```
RichStone, line 13: The previous version I copied this from had an if check for rails server running here, not sure this is needed (it was also erroring out for me 🤷).

config/cable.yml
```diff
@@ -6,7 +6,7 @@ test:
 
 production:
   adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: "<%= ENV.fetch('REDIS_URL') { ENV["KAMAL_SETUP"] == "true" ? "redis://:#{ENV['REDIS_PASSWORD']}@#{ENV['REDIS_HOST']}:#{ENV['REDIS_PORT']}/1" : "redis://localhost:6379/1" } %>"
```
RichStone, line 9:
```suggestion
  url: "<%= ENV.fetch('REDIS_URL') { "redis://localhost:6379/1" } %>"
```

I haven't found a way to re-use local vars inside the deploy.yml, so ended up hard-hacking this here. But probably we should just hardcode the REDIS_URL in the deploy.yml.

Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+
+# Copy built artifacts: gems, application
+COPY --from=build /usr/local/bundle /usr/local/bundle
+COPY --from=build /rails /rails
+
+# Run and own only the runtime files as a non-root user for security
+RUN useradd rails --create-home --shell /bin/bash && \
+    mkdir -p /rails/storage && \
+    mkdir -p /rails/tmp/pids && \
+    chown -R rails:rails db log storage config/credentials/* tmp tmp/pids # This step will fail if you haven't done `EDITOR="code --wait" bin/rails credentials:edit --environment production` before.
```
RichStone, line 83:
```suggestion
    chown -R rails:rails db log storage tmp tmp/pids
```

I think this will solve it and we don't really need to encourage using Rails credentials?

config/database.yml
```diff
@@ -91,3 +91,5 @@ production:
   username: untitled_application
   password: <%= ENV['UNTITLED_APPLICATION_DATABASE_PASSWORD'] %>
   url: <%= ENV['DATABASE_URL'] %>
+#  host: <%#= ENV['DB_HOST'] %> # For Kamal deployments *instead of url*.
+#  port: <%#= ENV['DB_PORT'] %> # For Kamal deployments *instead of url*.
```
RichStone, line 95: We probably should hardcode a DATABASE_URL in the deploy.yml instead. 🤔

config/initializers/sidekiq.rb
```diff
@@ -6,4 +6,16 @@
   Sidekiq.configure_client do |config|
     config.redis = {ssl_params: {verify_mode: OpenSSL::SSL::VERIFY_NONE}}
   end
+elsif ENV["KAMAL_SETUP"] == "true"
```
RichStone, line 9: What would be a good place for a global `kamal?` helper?

config/deploy.yml
RichStone, line 1: All the FIXMEs currently need to be addressed manually after `bin/configure`.

config/deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+proxy:
+  ssl: true
+  host: app.yoursaas-fix-me.com # FIXME: The domain you want to use for your app. You need to create a CNAME record that points to the IP address of the server you can SSH into.
+  app_port: 3000
+
+# Credentials for your image host.
+registry:
+  # Specify the registry server, if you're not using Docker Hub
+  # server: registry.digitalocean.com / ghcr.io / ...
+  username: docker-hub-username # FIXME: The username you use to log into Docker Hub.
```
jagthedrummer, line 40: Since `username` and `password` are paired maybe they should both go in `secrets`? Related: Will **everyone** that works on a given project **always** use the same `username`? If not, then it should definitely not be hard-coded into something checked into the repo.

config/deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+    - UNTITLED_APPLICATION_DATABASE_PASSWORD
+    - REDIS_PASSWORD
+    - POSTGRES_PASSWORD
+  clear:
+    DB_HOST: untitled-application-db
+    POSTGRES_USER: untitled_application
+    POSTGRES_DB: untitled_application_production
+    DB_PORT: 5432
+    REDIS_HOST: untitled-application-redis
+    REDIS_PORT: 6379
```
jagthedrummer, line 66: I think some of these are likely to be things that you'd want to be different between staging and production (for example) and so we may not want to encourage them to be hard coded. And if we go with `REDIS_URL` and `DATABASE_URL` then a lot of these should just go away.

config/deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+    image: postgres:14
+    host: 999.999.999.999 # FIXME: The IP address of the server you can SSH into.
+    port: "127.0.0.1:5432:5432" # Not exposing to Internet.
+    env:
+      clear:
+        POSTGRES_USER: untitled_application
+        POSTGRES_DB: untitled_application_production
+      secret:
+        - POSTGRES_PASSWORD
+        - UNTITLED_APPLICATION_DATABASE_PASSWORD
```
jagthedrummer, line 94: More stuff that may be better handled as secrets and collapsed into a single `URL` setting.

config/initializers/sidekiq.rb
jagthedrummer: Probably here: https://github.com/bullet-train-co/bullet_train-core/blob/520133dab93d1b549c22f128bba190e5b9bf8516/bullet_train/lib/bullet_train.rb

config/initializers/sidekiq.rb
```diff
@@ -6,4 +6,16 @@
...
+    config.redis = {
+      url: ENV.fetch("REDIS_URL") { "redis://:#{ENV["REDIS_PASSWORD"]}@#{ENV["REDIS_HOST"]}:#{ENV["REDIS_PORT"]}" }
+    }
+  end
+
+  Sidekiq.configure_client do |config|
+    config.redis = {
+      url: ENV.fetch("REDIS_URL") { "redis://:#{ENV["REDIS_PASSWORD"]}@#{ENV["REDIS_HOST"]}:#{ENV["REDIS_PORT"]}" }
+    }
+  end
```
jagthedrummer, line 20: Sidekiq automatically looks for `REDIS_URL`, so if we use that this entire block becomes unnecessary.

config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
+  # Add /up route which is needed by Kamal.
+  get "/up", to: proc { [200, {}, ["OK"]] }
```
jagthedrummer, line 3: Maybe this should go in one of the routes files in `core`? I'm not sure... 🤔

Dockerfile
RichStone: Most probably yes, but then it's an issue of _existing_ apps, they'll need to fix it on their side. Would you agree that we should optimize for our defaults and the _starter_ repo?

Dockerfile
RichStone: Very valid point. We could make this a Dockerfile.kamal.

.kamal/secrets
```diff
@@ -0,0 +1,30 @@
...
+# POSTMARK_API_TOKEN=$(kamal secrets extract POSTMARK_API_TOKEN $SECRETS)
+
+# FIXME: Pick your strategy here:
+# INFO: With the current BT bin/configure setup, you need both vars with the same Postgres password.
+KAMAL_REGISTRY_PASSWORD=
+RAILS_MASTER_KEY=
+UNTITLED_APPLICATION_DATABASE_PASSWORD=
+POSTGRES_PASSWORD=
+REDIS_PASSWORD=
+POSTMARK_API_TOKEN=
```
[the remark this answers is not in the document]
RichStone: This needs to be version-controlled because it's part of the minimal `kamal init` setup and when you run commands like `kamal deploy` its default is to only consider files that are git-committed.

The only thing that shouldn't happen is that people have the secrets in clear text, thus the comments and different mechanisms kamal offers to mitigate that as described in the comments.

bin/configure-scripts/deploy_kamal_instructions.rb
RichStone: Yeah, that would be the ideal experience. Though, this feels like quite some extra work and we don't know whether Kamal will be the go-to option for the time to come. But yeah, the instructions suck. 😅

RichStone (commented):
@jagthedrummer, big thanks for giving it a thorough look, I'll give this another round of iteration.

The only thing that feels like a lot of extra effort compared to the possible ROI is making the configure experience awesome. I would love to do this, but we don't really know how much traction this deployment path will have in the future. 🤔

config/deploy.yml
RichStone: I'm not sure we'll be able to replace all of them with their `_URL` counterparts. _I think_ some of these vars could be default names that are used inside the containers.

RichStone: It might be fine to "hardcode" them here, because you'd typically create something like a new `deploy.staging.yml` to deploy to a different env. 🤔

RichStone: Now I remember, the problem I'm running into with these is that the URLs are "secrets" because they contain the password. So you'd define them in your local `.zshrc` (or similar) or 1Password, but those places have no knowledge of ports, hosts, etc.

```bash
export REDIS_URL="redis://:${REDIS_PASSWORD}@${REDIS_HOST}:${REDIS_PORT}/0"
export DATABASE_URL="postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${DB_HOST}:${DB_PORT}/${POSTGRES_DB}"
```

Then we'd have a bad type of duplication, where we'd need to fix things in both places (risking to forget it in one of them).

RichStone: We could probably put all of that into the secret env 🤔

Dockerfile
jagthedrummer: Yeah, we should definitely optimize for the defaults we ship in the starter repo. And maybe use fenced sections like we do in some of the other files to indicate where people should put things that they're likely to want to add. Stuff like native dependencies and what not.

.kamal/secrets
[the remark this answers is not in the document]
jagthedrummer: Ah, I see.

config/deploy.yml
jagthedrummer: Oof, I see what you mean. 🤔

jagthedrummer:
Thanks for the responses, @RichStone. FYI, I'm going to be pretty slammed for the next few days, so it'll probably be next week before I can get back to this.

RichStone:
> Thanks for the responses, @RichStone. FYI, I'm going to be pretty slammed for the next few days, so it'll probably be next week before I can get back to this.

@jagthedrummer, no worries, there are still quite some things left to fix to make this a great experience. I will dedicate a weekly effort to it, but it won't be my main focus now.

.kamal/secrets
```diff
@@ -0,0 +1,30 @@
+# The easiest way to get started here is to set your secrets as environment
+# variables in your bashrc/zshrc and add them here as values:
+#
+# KAMAL_REGISTRY_PASSWORD=$KAMAL_REGISTRY_PASSWORD
+# RAILS_MASTER_KEY=$(cat config/credentials/production.key)
```
RichStone, line 5:
```suggestion
# RAILS_MASTER_KEY=$(cat config/master.key)
```

Dockerfile
RichStone: I'm using this setup to deploy one of my apps and I tried to do it without the Rails credentials, but got an error, so Kamal might require them or I missed a spot to fully remove it, so this needs more looking into.

Dockerfile
```diff
@@ -0,0 +1,96 @@
+# syntax = docker/dockerfile:1
+ARG RUBY_VERSION=3.4.1
```
RichStone, line 2: I think all the versions (Ruby, node, yarn) here need to be synced with what we have on local, so they need to come from something like a global var or `.ruby-version`, `.node-version` (not sure what it would be for yarn - the version is just a string in the package.json atm).

RichStone:
More TODOs here are CI and showcasing how this works with an external password manager vault to avoid conflicts with your system's env vars.

Dockerfile
lcorneliussen: I have done a couple of attempts to pull the version from .ruby-version, but couldn't get it working from within the Dockerfile... Args don't seem to support shell expansions.

But most probably there is a kamal way to dynamically set build args:
https://kamal-deploy.org/docs/configuration/builder-examples/#configuring-build-args-for-new-images

ronald2wing:
This has been so long, even Laravel Saas is getting their own Kamal soon https://github.com/thedevdojo/wave/pull/234
Thank you for your work, by the way.

jagthedrummer:
@RichStone sorry for the delay here. In [this issue about docker](https://github.com/bullet-train-co/bullet_train/issues/2012) @jeff-french raised the possibility of BT providing a docker image instead of just a `Dockerfile`. I think if we did that then the `Dockerfile` for the starter repo might be as simple as this:

```
FROM bullet_train/bullet_train:v1.22

# Add your additional dependencies and customizations here

COPY Gemfile Gemfile.lock .

RUN bundle install --quiet --jobs 4

COPY . .

CMD ["sh", "-c", "./bin/rails server"]
```

I'm curious if that seems better or worse to you, as someone actively using docker. And/or maybe you know of reasons that it wouldn't actually end up being that simple.

RichStone:
> @RichStone sorry for the delay here. In [this issue about docker](https://github.com/bullet-train-co/bullet_train/issues/2012) @jeff-french raised the possibility of BT providing a docker image instead of just a `Dockerfile`. I think if we did that then the `Dockerfile` for the starter repo might be as simple as this:
> 
> ```
> FROM bullet_train/bullet_train:v1.22
> 
> # Add your additional dependencies and customizations here
> 
> COPY Gemfile Gemfile.lock .
> 
> RUN bundle install --quiet --jobs 4
> 
> COPY . .
> 
> CMD ["sh", "-c", "./bin/rails server"]
> ```
> 
> I'm curious if that seems better or worse to you, as someone actively using docker. And/or maybe you know of reasons that it wouldn't actually end up being that simple.

That would be the dream scenario, I struggled quite a bit with the Dockerfile and I think it's still full of redundancies (one of those "still needs cleanup" items in this PR). My only question would be about overrideability/configurability. For example, for the second app on the same server that I'm using the approach in this PR, I expose a different port at the end of the Dockerfile: instead of `EXPOSE 3000`, I do `EXPOSE 3003`.

Question is, can I just override it at the end of the file? Are there other things folks might want to be configurable in the base image and some how those options could be passed down?

jagthedrummer:
> Question is, can I just override it at the end of the file? Are there other things folks might want to be configurable in the base image and some how those options could be passed down?

These are good questions @RichStone. 🤔 We definitely want that kind of thing to be configurable, and ideally I think we wouldn't want to lock people to those kinds of configuration things. Maybe there's a case to be made that that sort of thing should go in the `Dockerfile` in the starter repo and that the image itself should be mainly focused on providing the run time environment and all the required dependencies.

jeff-french:
The general idea of this approach is that, yes, you can override or add more dependencies / configuration in your downstream Dockerfile. The upstream image will be focused on providing a "batteries included" runtime environment for the starter repo and allow downstream consumers to build on top of that.

`EXPOSE` is mostly for documentation and enabling some shortcuts. It doesn't *actually* expose anything unless you supply the `-P` flag to your `docker run` command. If you do, it automatically publishes and maps any ports defined by `EXPOSE` from the Dockerfile as well as from the Dockerfile of any base images. This can be handy, but is not really considered a great practice. Typically you'd want to be more intentional about which ports you are publishing and which port they are mapped to on the host (e.g. `docker run -d -p HOST_PORT:CONTAINER_PORT nginx`). If you are using a container orchestrator like Docker Compose, Kubernetes, etc, this would generally be handled at that level.

With all of that in mind, I would expect that the base image published by Bullet Train would not expose any ports and the Dockerfile in the repo *may* have an `EXPOSE 3000` or similar if anything.

RichStone:
> With all of that in mind, I would expect that the base image published by Bullet Train would not expose any ports and the Dockerfile in the repo _may_ have an `EXPOSE 3000` or similar if anything.

Thanks a lot for the explanation, this sounds like the way to go! 👍

RichStone:
Noting another TODO for me from a comment from someone who recently successfully went through the flow from the description:

> The only thing I had to add was libyaml-dev to the docker file. rest was good.

jagthedrummer:
@RichStone when you have a chance could you rebase this against `main` and/or replace the `Dockerfile` in this one with the one in `main`, and make any additional changes that we need to make there to support kamal?

PR #2005, opened 2025-03-21
remove deprec rails-ujs

yshmarov:
```diff
# old
- link_to "Delete", product_path(product), method: :delete
# new
+ link_to "Delete", product_path(product), turbo_method: :delete 
# best
button_to "Delete", product_path(product), method: :delete 
```

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -3,7 +3,6 @@
 // includes should be specified at the end of the file, not in this section. This helps avoid merge conflicts in the
 // future should the framework defaults change.
 
-import Rails from "@rails/ujs"
 import * as ActiveStorage from "@rails/activestorage"
 import "@hotwired/turbo-rails"
 import "./controllers"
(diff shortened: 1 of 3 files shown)


app/javascript/application.js
```diff
@@ -3,7 +3,6 @@
 // includes should be specified at the end of the file, not in this section. This helps avoid merge conflicts in the
 // future should the framework defaults change.
 
-import Rails from "@rails/ujs"
```
jagthedrummer, line 6: @yshmarov I think that removing this line would cause the `Rails.start()` call below (on line 12/11) to fail. Do you need to also remove that?

jagthedrummer:
Also this has broken a lot of tests.

app/javascript/application.js
yshmarov: If we are removing `rails/ujs` from `package.json`, we won't have a place to "import Rails" from

I can have a look at how/whether `Rails.start()` needs to be updated

jagthedrummer:
Closing in favor of: https://github.com/bullet-train-co/bullet_train/pull/2071

This is really a two part PR where you need to have it connected to a branch in the `core` repo so that the turbo fixes happen at the same time.

Issue #2011, opened 2025-03-25
Generating a Stimulus controller clobbers our custom controller loading code


jagthedrummer:
If you use the rails generator to generate a new stimulus controller it completely clobbers `app/javascript/controllers/index.js`. And included in the clobbering is a comment that says that file is meant to be an auto-generated file that you shouldn't change manually.

The file starts like this:

https://github.com/bullet-train-co/bullet_train/blob/103b00f0393e9caaa4ed73a18dde95b68182c379/app/javascript/controllers/index.js#L1-L50

And then if you run:

```
rails g stimulus MyNewController
```

You'll see this output:

```
      create  app/javascript/controllers/my_new_controller.js
       rails  stimulus:manifest:update
```

And then `app/javascript/controllers/index.js` will look like this:

```javascript
// This file is auto-generated by ./bin/rails stimulus:manifest:update
// Run that command whenever you add a new controller or create them with
// ./bin/rails generate stimulus controllerName

import { application } from "./application"

import MyNewController from "./my_new_controller"
application.register("my-new", MyNewController)
```

I think the options here are:

* Move our controller loading code somewhere other than `app/javascript/controllers/index.js` to allow the normal Rails Way™ of loading controllers to work.
* Override the `rails g stimulus` generator to _only_ generate a new controller and to skip calling the `stimulus:manifest:update` task.
* Override the `stimulus:manifest:update` task to not clobber `app/javascript/controllers/index.js` and probably have it output a warning message that points you to some sort of documentation about what's going on.

@pascallaliberte, do you have any other ideas? Or a feel for what would be the best way to resolve this?

pascallaliberte:
@jagthedrummer if there's something we can do to override the generator locally to skip the `stimulus:manifest:update` task, that would be best.

See related:
- https://github.com/bullet-train-co/bullet_train/issues/402

PR #2068, opened 2025-05-22
Avoid clobbering our Stimulus manifest file when generating a new controller

jagthedrummer:
Fixes https://github.com/bullet-train-co/bullet_train/issues/2011

With this task override active you'll see this when generating a new stimulus controller:

```
$ rails g stimulus MyNewController
      create  app/javascript/controllers/my_new_controller.js
       rails  stimulus:manifest:update
-----------------------------------------------------------------------------------
We are skipping the stimulus:manifest:update task to avoid clobbering our manifest.
If you need to run this task pass it the `--clobber` option like this:
    bin/rails stimulus:manifest:update -- --clobber
-----------------------------------------------------------------------------------
```

If you do `bin/rails stimulus:manifest:update -- --clobber` then we avoid overriding the task, and we run the original task, which will clobber the file we're trying to protect.

### DIFF
--- a/lib/tasks/stimulus_manifest_update_override.rake
+++ b/lib/tasks/stimulus_manifest_update_override.rake
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-#
+
 # We clear and override the stimulus:manifest:update task to avoid having it clobber
 # our custom and non-standard stimulus manifest file.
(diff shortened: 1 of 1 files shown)


pascallaliberte (approved):
👌 this is excellent

PR #2089, opened 2025-06-03
Updates for `pagy` 9

jagthedrummer:
Joint PR: https://github.com/bullet-train-co/bullet_train-core/pull/1150

The test failures here are in the part of the test suite that's testing against the published gems, which expect `pagy` version 8, so they're to be expected.


RichStone (approved):
Thanks for the careful doc comments updates!

PR #2092, opened 2025-06-04
Webhook deactivation system

Riddlerrr:
This is a paired PR with the PR to BT core: https://github.com/bullet-train-co/bullet_train-core/pull/1151
It only makes sense when they are both included in the same release.

### What is included in this PR

- DB migrations are needed to add the missing fields required for [the PR](https://github.com/bullet-train-co/bullet_train-core/pull/1151)
- DB migrations that add missing indexes for webhooks tables to improve performance 🚅
- Add the sidekiq-scheduler gem and schedule the Webhooks::Outgoing::EndpointsDeactivationJob to run every 6 hours.
- Mailer notification and translation templates

### Important DB migrations

This PR includes two migrations that add missing indexes for the webhooks_outgoing_deliveries and webhooks_outgoing_delivery_attempts tables. These tables can be large for existing projects that heavily use webhooks, so you need to be careful with the migrations. 

In the current implementation, adding indexes occurs with concurrency applied, so it should go smoothly. But you might do it your own way.

Migrations:
- `db/migrate/20250528140801_add_index_for_webhooks_outgoing_deliveries.rb`
- `db/migrate/20250528141219_add_index_for_webhooks_outgoing_attempts.rb`

### DIFF
--- a/db/migrate/20250528140801_add_index_for_webhooks_outgoing_deliveries.rb
+++ b/db/migrate/20250528140801_add_index_for_webhooks_outgoing_deliveries.rb
@@ -3,6 +3,6 @@ class AddIndexForWebhooksOutgoingDeliveries < ActiveRecord::Migration[8.0]
 
   def change
     add_index :webhooks_outgoing_deliveries, [:endpoint_id, :event_id], algorithm: :concurrently
-    add_index :webhooks_outgoing_deliveries, [:delivered_at], algorithm: :concurrently
+    add_index :webhooks_outgoing_deliveries, [:delivered_at, :created_at], algorithm: :concurrently
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -302,7 +302,7 @@ ActiveRecord::Schema[8.0].define(version: 2025_05_28_141219) do
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
     t.datetime "delivered_at", precision: nil
-    t.index ["delivered_at"], name: "index_webhooks_outgoing_deliveries_on_delivered_at"
+    t.index ["delivered_at", "created_at"], name: "idx_on_delivered_at_created_at_5c26227119"
     t.index ["endpoint_id", "event_id"], name: "index_webhooks_outgoing_deliveries_on_endpoint_id_and_event_id"
   end
 


db/migrate/20250526095219_add_deactivation_fields_to_endpoints.rb
```diff
@@ -0,0 +1,9 @@
+class AddDeactivationFieldsToEndpoints < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
```
RichStone, line 2: Do we need this in all migration files? Because of concurrent index additions but why do we need the concurrency? (just curious)

test/mailers/previews/webhooks/outgoing/endpoint_mailer_preview.rb
```diff
@@ -0,0 +1,23 @@
+class Webhooks::Outgoing::EndpointMailerPreview < ActionMailer::Preview
+  def deactivation_limit_reached
+    endpoint = Webhooks::Outgoing::Endpoint.last ||
+      Webhooks::Outgoing::Endpoint.new(
```
RichStone, line 4: We'll notify about the last or new endpoint? 🤔 

How is it guaranteed that the last endpoint is the one failing?

`new` doesn't make sense to me at all, not sure why it's there. 

Imo, it should have the endpoint passed to it in the params?

Same for the `deactivated` method below.

config/locales/en/webhooks/outgoing/endpoint_mailer.en.yml
```diff
@@ -0,0 +1,93 @@
...
+                        </td>
+                      </tr>
+                    </table>
+                  </td>
+                </tr>
+              </table>
+          signature: &SIGNATURE
+            html:
+              <p>If you have any questions, please don't hesitate to <a href="mailto:%{support_email}">send us an email</a>.</p>
+              <p>Thanks,<br>The Untitled Application Team</p>
```
RichStone, line 50: IIRC you'll need to configure something (the file name) in the `bin/configure` script so this gets overridden with app name when the customer starts the repo. I might be mistaken, but something to check for ya.

app/views/webhooks/outgoing/endpoint_mailer/deactivation_limit_reached.html.erb
RichStone, line 1: I remember we discussed this, but looking at this again, I'm even less sure why we need the mailers to be in the starter repo. 😅 
I think the regular BT thing to do would be to put them into core and just have the translations in the starter repo. (see how this works for example with the UserMailer)

RichStone (commented):
Leaving some comments for you to defend before I approve. 😁 ⚔️

RichStone:
Failing CI about openapi schema seems unrelated.

RichStone:
> Failing CI about openapi schema seems unrelated.

Ah, I see now, it might be a missing translation:

```
SuperScaffoldingOpenAPITest
2025-06-04 11:06:39 +0000 Rack app ("GET /api/v1/openapi.yaml" - ([127](https://github.com/bullet-train-co/bullet_train/actions/runs/15440599674/job/43457277594?pr=2092#step:18:128).0.0.1)): #<ActionView::Template::Error: Translation missing: en.webhooks/outgoing/endpoints.fields.deactivation_limit_reached_at.heading>
validating http://127.0.0.1:3003/api/v1/openapi.yaml...
Failed to resolve API description at http://127.0.0.1:3003/api/v1/openapi.yaml:

  - Failed to load http://127.0.0.1:3003/api/v1/openapi.yaml: 500 Internal Server Error
```

app/views/webhooks/outgoing/endpoint_mailer/deactivation_limit_reached.html.erb
Riddlerrr: Yes, I saw how it works for UserMailer.
For me, I'd want to have it in the template repository so that I can easily find and change it.
But of course I can easily move it to the core gem.

@jagthedrummer WDYT? Should I move the email templates to the core gem or leave them here?

jagthedrummer: @Riddlerrr I think absolutely as much as possible should go in `core`. The idea is that if someone has not consciously and explicitly made the choice to deviate from BT defaults then we want stuff to be "hidden" from them so that they don't accidentally get themselves into hairy merge-conflict territory as we make changes upstream. Even the translations should go there I think. (I'm not sure why some translations are in the starter repo and some are in `core` but we definitely rely on some that are part of `core` so we know that pattern works.)

config/sidekiq.yml
```diff
@@ -9,3 +9,10 @@ production:
   - mailers
   - low
   - action_mailbox_routing
+:scheduler:
+  webhooks_endpoints_deactivation:
+    enabled: true
+    interval:
+      - 6h
+      - first_in: 5m
+    class: Webhooks::Outgoing::EndpointsDeactivationJob
```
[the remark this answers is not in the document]
Riddlerrr: It looks like we don't need to comment this out. They're just unused keys without the schedule gem. They don't do anything or break anything.

db/migrate/20250526095219_add_deactivation_fields_to_endpoints.rb
Riddlerrr: I've added a section to the PR's description describing the migrations with indexes

config/sidekiq.yml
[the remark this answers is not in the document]
jagthedrummer: Excellent! I had hoped that would be the case.

test/system/pagination_test.rb
```diff
@@ -1,4 +1,5 @@
 require "application_system_test_case"
+require "pagy"
```
jagthedrummer, line 2: I don't think we should need this...

Riddlerrr:
Closing this in favor of https://github.com/bullet-train-co/bullet_train/pull/2114

PR #2190, opened 2025-09-02
Exit after showing error message if bin/theme is missing arguments

jagthedrummer:
Previously we'd show an error about missing argument, but then try to continue running the script, which would then throw a different less descriptive error, which makes it easy to miss the descriptive one that's shown earlier.


jagthedrummer:
On second thought, I'm not sure if this is going to change the way in which things break in ways that aren't helpful. Specifically we rely on `bin/theme` outputting a useful string, and with the old behavior it would actually raise an error after outputting a message. But with this change it would not raise that error.

PR #2222, opened 2025-09-25
DO NOT MERGE - New sortable controller proof of concept

jagthedrummer:
This is a first pass at a sortable controller that doesn't rely on `dragula`, which is pretty much abandoned, and forces us to include several other dependencies. Related issue: https://github.com/bullet-train-co/bullet_train-core/issues/1215

It also addresses https://github.com/bullet-train-co/bullet_train-core/issues/826 by adding a drag handle, adding some better styling for in-process drags, and disabling dragging on other parts of the row so that text in the row is still selectable.

Dragging and dropping is working reasonably well in Firefox, Safari and Chrome, but it does look slightly different in each browser.

### Firefox:

CleanShot 2025-09-26 at 14 55 28

### Safari:

CleanShot 2025-09-26 at 14 57 17

### Chrome:

CleanShot 2025-09-26 at 14 59 34

This demo PR consists of two commits, one to generate the old sortable super scaffolding and [this commit](https://github.com/bullet-train-co/bullet_train/pull/2222/commits/340ebe9f70c338b6b1f88af1cac93349c788a365) which contains a `new-sortable` stimulus controller and some updates to the templates for the super scaffolded demo model.

In order to get this into Bullet Train we'd need to do the following:

* move the Stimulus controller into `core`
* update super scaffolding in `core` to add the new `th` and `td` columns that we need for drag handles
* document any changes that people would need to make to upgrade to the new controller
* remove `dragula` as a dependency in all the appropriate spots in `core`
* document how people could continue to use the old `dragula` based controller
* update the controller to emit `sortable:*` events [as described here](https://bullettrain.co/docs/super-scaffolding/sortable#events)

### DIFF
--- a/app/views/account/pages/_index.html.erb
+++ b/app/views/account/pages/_index.html.erb
@@ -22,6 +22,7 @@
           <table class="table">
             <thead>
               <tr>
+                <th class="w-6"></th>
                 <%= render "shared/tables/select_all" %>
                 <th><%= t('.fields.name.heading') %></th>
                 <%# 🚅 super scaffolding will insert new field headers above this line. %>
@@ -29,7 +30,7 @@
                 <th class="text-right"></th>
               </tr>
             </thead>
-            <tbody data-controller="sortable" data-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>">
+            <tbody data-controller="new-sortable" data-new-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>">
               <%= render partial: 'account/pages/page', collection: pages %>
             </tbody>
           </table>
--- a/app/views/account/pages/_index.html.erb
+++ b/app/views/account/pages/_index.html.erb
@@ -1,0 +1,62 @@
+<% pages = pages.accessible_by(current_ability) %>
+<% team = @team %>
+<% context ||= team %>
+<% collection ||= :pages %>
+<% hide_actions ||= false %>
+<% hide_back ||= false %>
+
+<% pagy ||= nil %>
+<% pagy, pages = pagy(pages, page_param: :pages_page) unless pagy %>
+
+<%= action_model_select_controller do %>
+  <% cable_ready_updates_for context, collection do %>
+    <%= render 'account/shared/box', pagy: pagy do |box| %>
+      <% box.title t(".contexts.#{context.class.name.underscore}.header") %>
+      <% box.description do %>
+        <%= t(".contexts.#{context.class.name.underscore}.description#{"_empty" unless pages.any?}") %>
+        <%= render "shared/limits/index", model: pages.model %>
+      <% end %>
+
+      <% box.table do %>
+        <% if pages.any? %>
+          <table class="table">
+            <thead>
+              <tr>
+                <%= render "shared/tables/select_all" %>
+                <th><%= t('.fields.name.heading') %></th>
+                <%# 🚅 super scaffolding will insert new field headers above this line. %>
+                <th><%= t('.fields.created_at.heading') %></th>
+                <th class="text-right"></th>
+              </tr>
+            </thead>
+            <tbody data-controller="sortable" data-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>">
+              <%= render partial: 'account/pages/page', collection: pages %>
+            </tbody>
+          </table>
+        <% end %>
+      <% end %>
+
+      <% box.actions do %>
+        <% unless hide_actions %>
+          <% if context == team %>
+            <% if can? :create, Page.new(team: team) %>
+              <%= link_to t('.buttons.new'), [:new, :account, team, :page], class: "#{first_button_primary(:page)} new" %>
+            <% end %>
+          <% end %>
+
+          <%# 🚅 super scaffolding will insert new targets one parent action model buttons above this line. %>
+          <%# 🚅 super scaffolding will insert new bulk action model buttons above this line. %>
+          <%= render "shared/bulk_action_select" if pages.any? %>
+
+          <% unless hide_back %>
+            <%= link_to t('global.buttons.back'), [:account, context], class: "#{first_button_primary(:page)} back" %>
+          <% end %>
+        <% end %>
+      <% end %>
+
+      <% box.raw_footer do %>
+        <%# 🚅 super scaffolding will insert new action model index views above this line. %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/account/pages/_index.html.erb
+++ b/app/views/account/pages/_index.html.erb
@@ -19,10 +19,10 @@
 
       <% box.table do %>
         <% if pages.any? %>
-          <table class="table">
+          <table class="table" data-controller="confirm-reorder">
             <thead>
               <tr>
-                <th class="w-6"></th>
+                <!--<th class="w-6"></th>-->
                 <%= render "shared/tables/select_all" %>
                 <th><%= t('.fields.name.heading') %></th>
                 <%# 🚅 super scaffolding will insert new field headers above this line. %>
@@ -30,7 +30,9 @@
                 <th class="text-right"></th>
               </tr>
             </thead>
-            <tbody data-controller="new-sortable" data-new-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>">
+            <tbody data-controller="new-sortable" data-new-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>"
+              data-action="new-sortable:drop->confirm-reorder#requestConfirmation"
+              >
               <%= render partial: 'account/pages/page', collection: pages %>
             </tbody>
           </table>
--- a/app/avo/resources/page.rb
+++ b/app/avo/resources/page.rb
@@ -4,7 +4,7 @@ class Avo::Resources::Page < Avo::BaseResource
   # self.search = {
   #   query: -> { query.ransack(id_eq: q, m: "or").result(distinct: false) }
   # }
-  
+
   def fields
     field :id, as: :id
     field :team, as: :belongs_to
--- a/app/avo/resources/page.rb
+++ b/app/avo/resources/page.rb
@@ -1,0 +1,14 @@
+class Avo::Resources::Page < Avo::BaseResource
+  # self.includes = []
+  # self.attachments = []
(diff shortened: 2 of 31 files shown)


app/views/account/pages/_index.html.erb
```diff
@@ -22,14 +22,15 @@
...
+                <th class="w-6"></th>
                 <%= render "shared/tables/select_all" %>
                 <th><%= t('.fields.name.heading') %></th>
                 <%# 🚅 super scaffolding will insert new field headers above this line. %>
                 <th><%= t('.fields.created_at.heading') %></th>
                 <th class="text-right"></th>
               </tr>
             </thead>
-            <tbody data-controller="sortable" data-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>">
+            <tbody data-controller="new-sortable" data-new-sortable-reorder-path-value="<%= url_for [:reorder, :account, context, collection] %>">
```
jagthedrummer, line 33: `new-sortable` isn't a great name. If we can be backwards compatible enough I'd love to move the old `dragula` based controller to `dragula-sortable` or something and make the new version `sortable`. That would make for the most seamless upgrade for people.

pascallaliberte:
Oh I like where this is going. Removing the `dragula` dependency completely.

## A few notes:

1. I'd like to see the bulk of the event handling, injection and removal be placed into an imported package that we host, say in `support/`. On `connect()`, we'd `this.plugin = Draggable.new(...)` and on `disconnect()` we'd `this.plugin.remove()`
2. I like the default classes. This might be a good candidate for a new `draggable.css` with just some `.draggable-active-dropzone`, etc. classes with some `@apply`.
3. I like the look! Big improvement
4. I wonder about handles. More below.

## About handles:

I'm concerned about the migration for existing tables. I'd use `ti-line-double` for the icon at least, but even then.

For now, I think we should instead continue to go handle-less for `sortable` tables, and keep the existing drag-the-whole-row functionality.

Later, I'd do the following: we could introduce a `sortable-with-toggle` flag or something, which instead of showing handles all the time, and instead of having the whole row be draggable, the `sortable-with-toggle` introduces a "Modify" button above tables. This makes handles appear, handles that are programmatically added by the sortable controller itself. If what's sortable is a table, we add on `connect()` a `th` in the `thead` and some `td` in the rows, hidden normally, but revealed on toggling the Modify button. Not for now, better UX.

jagthedrummer:
Thanks for giving it a look, @pascallaliberte!

Can you say more about what you see as being the advantages to having a separate `Draggable` class? Specifically, do you see it as improving _this_ use case in the immediate term, or is more of a "we might need to have other draggable things one day" situation? If it's the second thing, is there any reason not to delay extracting it until we have a second use case (to help us draw the lines in the abstraction at the right place)?

I had a similar thought about adding the drag handles via code instead of manually in the template, especially for making it easier to upgrade. Thanks for the suggestion of `ti-line-double`, I wasn't sure the best icon to use, but I do think that one is probably better than the double arrows.

I'm not sure that I'm a fan of adding a bunch of complexity around supporting different ways of sorting (enabling both "whole row" and "disappearing drag handles"). I think we should probably stick to the general Bullet Train ethos of figuring out a good workable default for most use cases and making that The Way. People can always customize further if they need to.

So, in an effort to dial in on what makes sense as a widely workable default here are some general thoughts.

* Making the whole row draggable without a handle is pretty poor UI/UX in my opinion for several reasons.
    * Highlighting/selecting text in the row is completely impossible.
    * Clicking on links/button in the row can be janky if you do even a slightly long click or even a pixel of movement during the click.
    * It's possible to accidentally reorder things (like if you try to highlight some text or accidentally long click a link/button).
    * There's no indication that the items are sortable, you have to just know. (https://github.com/bullet-train-co/bullet_train-core/issues/826)
* Adding a "modify order" button has some pros and cons.
    * Pro: It would explicitly draw attention to the fact that the items are sortable.
    * Pro: It would be much harder to accidentally reorder something. (Permanent handles would somewhat minimize the potential, but it would still be easier than needing to proactively get into "reorder mode".)
    * Con: We either have to make the upgrade process more involved by having people add the button, or we have to add it programmatically which would make it harder for people to customize as compared to the button living in a template.
    * Con: Adds more stuff (clutter?) to the UI.

So, given all of the above, I think my current preferences in order are:
1. Permanent drag handles that we add programmatically. (I think this is the sweet spot in tradeoffs between simplicity and easy upgrades.)
2. Disappearing drag handles with a "modify order" toggle. (I think this could potentially still make for easy upgrades, but I have concerns about how we'd programmatically add the button and how people could customize it.)

I'm generally against maintaining our current "invisibly sortable whole row" approach. I realize that that is the most "backwards compatible" in terms of it changing the behavior the least, but I just think the current behavior is bad and worth changing.

I'm certainly open to being convinced about other possibilities. Would love to hear your thoughts!

pascallaliberte:
Re: a `Draggable` class:

To me it's because of clarity of what the sortable_controller affords.

I see however that if we add an `import`, that messes up the possibly of using importmaps.

My suggestion then: have the Draggable class defined (not exported) below the sortable_controller's main code.

jagthedrummer:
@pascallaliberte I don't think I understand what you mean by this:

> To me it's because of clarity of what the sortable_controller affords.

Can you say more? Or maybe I need to see some example/pseudo code of how the two would fit together? My current impression is that it would add _complexity_ without adding clarity, but I assume that's because I'm not understanding what you're getting at.

Theoretically I don't think that adding an `import` should prevent moving to importmaps. Importmaps are supposed to make it easy to `import` random things without them needing to be bundled into the same physical file.

pascallaliberte:
@jagthedrummer 

> Can you say more? Or maybe I need to see some example/pseudo code of how the two would fit together? 

`sortable_controller` would have the following methods:

```js
connect() // inits the plugin, defines callback for when re-order occurred
saveOrder() // saves the updated order, called back from plugin
disconnect() // tears down the plugin
```

It's a clear contract.

The plugin itself, let's say called with signature `Draggable.new(config, reorderCallback) // returns instance`, would remove the handle* methods from the main controller, allow defining the selector maybe for finding the `closest()` up the event.target parents, etc.

Cleaner separation of concerns, and a pattern we use elsewhere (`initPlugin`, `teardownPlugin`)

jagthedrummer:
Oh, thank you @pascallaliberte! I think I see what you're getting at now. I think the idea of the plugin being called `Draggable` had me imagining something way more generalized than what it sounds like you're proposing.

Is the following a reasonable description? (Just putting it in my own words to make sure we're pointed in the same direction.)

The `Draggable` class you're talking about is a specific implementation of a sortable html table that _only_ handles the visual/DOM side of things, and then the `SortableController` would instantiate a sortable html table and would be notified after a `drop` so that it could save the order.

`Draggable` == generic sortable html table, with automatic handles, etc...
`SortableController` == sortable list of Bullet Train models

jagthedrummer:
OK, @pascallaliberte I think [this commit](https://github.com/bullet-train-co/bullet_train/pull/2222/commits/562c58b9ba37be4f5d73be1b48aa3c8b4c41c253) is a start on what you're talking about. I do think this is a definite improvement.

pascallaliberte:
`SortableTable` is a great name!

This is very nice.👍

jagthedrummer:
Good suggestion on making a separation!

pascallaliberte:
Regarding the injection of the handle, I'd recommend that we don't add a new column but instead inject the handle into the first cell of each row unless cellWithDragHandle targets are defined. We'd also only inject the handle on rows with the `data-id` param.

We'd adjust the padding-left of that cell, add relative class to the cell, and we'd inject an absolutely-positioned handle icon.

jagthedrummer:
@pascallaliberte I think you're right that we should check for `data-id` on a row before injecting the handle. I'm curious why you'd avoid adding a new cell and why you like absolute positioning for the handle.

pascallaliberte:
@jagthedrummer I thought of it more, and adding a column is the way to go.

I'm concerned with the column being added via Stimulus (after page paint) creating a janky experience. The best way to solve that would be to super scaffold that column. The other way, add it programmatically, but with a grow animation. The other: if a handle is missing, it's the whole row that's the handle (as before), with documentation on how to add the column to your _tangible_thing row partial for the handle-only experience.

jagthedrummer:
So far I haven't been able to get a grow animation to work on a table cell in a way that would actually prevent any jumping/janking of the page. It immediately claims the full space, and then just grows the content.

CleanShot 2025-09-30 at 12 41 57

jagthedrummer:
@pascallaliberte here's my latest thinking on the drag handle question. I'm hoping we can find a happy balance between easy upgrades and the ideal situation going forward.

1. For easing upgrades the `SortableController`/`SortableTable` will default to automatically adding a new cell with the drag handle (without any animation for now, unless we can get it to work as expected).
2. `SortableController`/`SortableTable` will look for a `data-sortable-add-drag-handles-value` attribute on the sortable element. If that is set to `false` then we won't add the handles.
3. We'll update the superscaffolding templates to include handles directly in the template along with setting `data-sortable-add-drag-handles-value` to `false`.
4. In the release notes (and maybe on the main sortable docs) we'll document what changes would need to be made to add drag handles to the template and disable the addition of automatic handles.

I think this will make it so that new apps and new supper scaffolds in existing apps will default to having handles in the template, but existing super scaffolds in existing apps will get automatic handles without needing to change anything.

Does that sound reasonable and workable to you?

pascallaliberte:
> 2. `SortableController`/`SortableTable` will look for a `data-sortable-add-drag-handles-value` attribute on the sortable element. If that is set to `false` then we won't add the handles.
> 3. We'll update the superscaffolding templates to include handles directly in the template along with setting `data-sortable-add-drag-handles-value` to `false`.

Here's another option:
* The handle element should have a `data-sortable-target="handle"`.
* The `SortableController` has a `get shouldAddHandles()` that checks whether `this.handleTargets.length > 0`
* Whether added by the scaffolding or via the Stimulus controller, we'll need to make sure that target is defined on the handle.
* This makes everything idempotent out of the box

No need for enter animations for now.

jagthedrummer:
Good call on using targets instead of a magic class name, @pascallaliberte. 

Here's my first pass at moving all of this stuff into `core` and updating the scaffolding transformer to do the right thing.

https://github.com/bullet-train-co/bullet_train-core/pull/1222

jagthedrummer:
Added this to `core` in https://github.com/bullet-train-co/bullet_train-core/pull/1222

PR #2269, opened 2025-10-29
image and video partial + making attachments in action text actually render

1MasterSpark:
added a partial for rendering images and videos
added _html.html.erb and changed _blob.html.erb so that they render attachments that are in ActionText


1MasterSpark:
Hi @jagthedrummer 

Thanks for your feedback. I guess I should have submitted the PR to bullet_train-core, not here. I'll close this PR now, and try to start another two, one to bullet_train-core, and one here wih the correct changes

PR #2293, opened 2025-11-06
making previews for attachments in action text

1MasterSpark:
In [this PR](https://github.com/bullet-train-co/bullet_train/pull/2269), I made it so that attachments in action text render previews for images, videos, GIFs, PDFs, and files that can't be rendered like ZIP files. This time I replaced all the hardcoded style attributes in _blob with Tailwind CSS, and I'm making two PRs, this one and [this one](https://github.com/bullet-train-co/bullet_train-core/pull/1266). @jagthedrummer what do you think?

### DIFF
--- a/app/views/active_storage/blobs/_blob.html.erb
+++ b/app/views/active_storage/blobs/_blob.html.erb
@@ -1,14 +1,35 @@
-<figure class="attachment attachment--<%= blob.representable? ? "preview" : "file" %> attachment--<%= blob.filename.extension %>">
-  <% if blob.representable? %>
-    <%= image_tag blob.representation(resize_to_limit: local_assigns[:in_gallery] ? [ 800, 600 ] : [ 1024, 768 ]) %>
+<% options ||= {} %>
+
+<% if blob.representable? %>
+  <%
+    image_source = nil
+    if blob.content_type == "application/pdf"
+      # If it's a PDF, increase the DPI to make the preview image readable
+      options[:width] ||= 400
+      options[:height] ||= 400
+      preview_options = { resize_to_limit: [options[:width], options[:height]] }
+      preview_options[:saver] = { dpi: 300 }
+      image_source = blob.representation(preview_options)
+    else
+      image_source = blob
+    end
+  %>
+
+  <%= link_to url_for(blob), target: "_blank", title: "View original: #{blob.filename}", class: "inline-block" do %>
+    <%= image_tag image_source, options.merge(class: "rounded-md max-w-full h-auto") %>
   <% end %>
 
-  <figcaption class="attachment__caption">
-    <% if caption = blob.try(:caption) %>
-      <%= caption %>
-    <% else %>
-      <span class="attachment__name"><%= blob.filename %></span>
-      <span class="attachment__size"><%= number_to_human_size blob.byte_size %></span>
+<% elsif blob.video? %>
+  <video controls width="<%= options[:width] %>" preload="metadata" class="max-w-full h-auto rounded-md">
+    <source src="<%= rails_blob_url(blob) %>" type="<%= blob.content_type %>">
+  </video>
+
+<% else %>
+  <%# Fallback for files that can't be represented as images, like zip files %>
+  <div class="p-3 border border-gray-300 rounded-md flex items-center space-x-2">
+    <%= link_to url_for(blob), target: "_blank", class: "flex items-center text-blue-600 hover:underline" do %>
+      <span class="text-xl mr-2">📄</span>
+      <span><%= blob.filename %></span>
     <% end %>
-  </figcaption>
-</figure>
+  </div>
+<% end %>
--- a/config/initializers/action_text_video_support.rb
+++ b/config/initializers/action_text_video_support.rb
@@ -1,0 +1,11 @@
+# This is to render previews for video attachments in Action Text
+# Source: https://mileswoodroffe.com/articles/action-text-video-support
+Rails.application.config.after_initialize do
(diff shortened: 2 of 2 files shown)


jagthedrummer:
Thanks for taking another pass at this, @1MasterSpark! This is definitely looking a lot closer. I'm going to plan to spend some time with it tomorrow testing out various combinations of asset handling to make sure this will cover all of the various cases that we need to support.

app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -1,14 +1,35 @@
...
+      preview_options[:saver] = { dpi: 300 }
+      image_source = blob.representation(preview_options)
+    else
+      image_source = blob
+    end
+  %>
+
+  <%= link_to url_for(blob), target: "_blank", title: "View original: #{blob.filename}", class: "inline-block" do %>
+    <%= image_tag image_source, options.merge(class: "rounded-md max-w-full h-auto") %>
   <% end %>
```
jagthedrummer: We're definitely going to need to do something different here. I think we'll still need to use `blob.representation` for _most_ media formats. Not calling `representation` means that we're linking directly to the uploaded assets instead of creating a thumbnail. For gifs, this is (probably) a good thing, but for huge jpegs it's definitely not good. We may want to give app developers some easy way to toggle whether or not we embed the full gif or a shot of the first frame. I think `representable?` might cover more media formats than _only_ images, so we'll definitely need `representation` for those.

app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -1,14 +1,35 @@
-<figure class="attachment attachment--<%= blob.representable? ? "preview" : "file" %> attachment--<%= blob.filename.extension %>">
```
jagthedrummer: We definitely need to keep this outer `figure` element for general backwards-compatibility friendliness to downstream developers who are trying to upgrade existing apps without springing unexpected changes on their users. (For example, people may have written custom CSS targeting these `figure` elements.)

app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -1,14 +1,35 @@
...
-  <figcaption class="attachment__caption">
-    <% if caption = blob.try(:caption) %>
-      <%= caption %>
-    <% else %>
-      <span class="attachment__name"><%= blob.filename %></span>
-      <span class="attachment__size"><%= number_to_human_size blob.byte_size %></span>
+<% elsif blob.video? %>
+  <video controls width="<%= options[:width] %>" preload="metadata" class="max-w-full h-auto rounded-md">
+    <source src="<%= rails_blob_url(blob) %>" type="<%= blob.content_type %>">
+  </video>
```
jagthedrummer: This seems reasonable. We'll need to double check that it works with `config.active_storage.service = :cloudinary`.

config/initializers/action_text_video_support.rb
```diff
@@ -0,0 +1,11 @@
+# This is to render previews for video attachments in Action Text
+# Source: https://mileswoodroffe.com/articles/action-text-video-support
+Rails.application.config.after_initialize do
+  default_allowed_attributes = Rails::HTML5::Sanitizer.safe_list_sanitizer.allowed_attributes + ActionText::Attachment::ATTRIBUTES.to_set
+  custom_allowed_attributes = Set.new(%w[controls])
+  ActionText::ContentHelper.allowed_attributes = (default_allowed_attributes + custom_allowed_attributes).freeze
+
+  default_allowed_tags = Rails::HTML5::Sanitizer.safe_list_sanitizer.allowed_tags + Set.new([ ActionText::Attachment.tag_name, "figure", "figcaption" ])
+  custom_allowed_tags = Set.new(%w[audio video source])
+  ActionText::ContentHelper.allowed_tags = (default_allowed_tags + custom_allowed_tags).freeze
```
jagthedrummer: I'm not sure we need any of this. [We handle santizing html in the `sanitize_html` method](https://github.com/bullet-train-co/bullet_train-core/blob/f966becdff7e12f1e394f26e3226658d643030a1/bullet_train-fields/app/helpers/fields/html_editor_helper.rb#L4-L15) and we already include `figure` and `figurecaption`. I suspect that this was necessary because the companion PR [removes the use of our `html_sanitize` method](https://github.com/bullet-train-co/bullet_train-core/pull/1266/files#diff-a851fda9025e6cc01b4f2d57da897b4db992da2b75d6e140c44ba00d11a81c75L8).

jagthedrummer (changes requested):
OK, @1MasterSpark I gave this a close reading just to help myself understand what particular problems you might have been looking to fix. In the process I left some comments to try to help us land in the right spot. [As I mentioned on the companion PR](https://github.com/bullet-train-co/bullet_train-core/pull/1266#issuecomment-3522932435) it might be good to handle each individual format that we want to fix as a separate issue so that we end up with a sequence of change that are each smaller and easier to grok.

I think I see several different conceptual things happening in these PRs, and I'm not sure we can accept all of them.

* ✅ make videos render with controls - We should definitely do this.
* ✅ make gifs render the animation instead of a still - We may want to give devs an easy way to toggle this, but probably default it to the new behavior.
* ❓ stop using `html_sanitize` - If there's a strong argument to be made for that we could certainly consider it as a stand-alone issue.
* ❌ change the structure of what we render in `blob` - We need to preserve existing behavior as much as possible so we don't unnecessarily break things for existing apps.
* ❌ standardize on `ActiveStorage` and The Way™️  - We still need to support Cloudinary.

Thanks again for working on this!

config/initializers/action_text_video_support.rb
1MasterSpark: Yeah I removed html_sanitize because it seemed to be filtering out a lot of attachments

PR #2299, opened 2025-11-11
Migration to fix mismatches between memberships.user_email and users.email

jagthedrummer:
Fixes lingering data inconsistencies caused by: https://github.com/bullet-train-co/bullet_train-core/issues/1268

Joint PR: https://github.com/bullet-train-co/bullet_train-core/pull/1269


pascallaliberte (approved):
I like the explanations

Issue #2303, opened 2025-11-12
GIFs in a `trix_editor` field aren't animated when displaying outside the editor


jagthedrummer:
If you add an animated GIF into a Trix editor it will display inside the editor as expected and the animation runs:

![Image](https://github.com/user-attachments/assets/3c48015d-a3f9-4034-beb0-8d805dca4e18)

But after you save the record we only show a still shot of the first frame of the GIF instead of showing the full animation:

![Image](https://github.com/user-attachments/assets/0f61510d-eb2d-433b-87a6-7282b235b580)

This is when using:

```ruby
config.active_storage.service = :local
```

And enabling the `image_processing` gem in `Gemfile`:

```ruby
gem "image_processing", "~> 1.2"
```

PR #2305, opened 2025-11-12
Allow gifs to animate when embedded in ActionText

jagthedrummer:
By default the `vips` library will read only the first frame of a `.gif` when creating a "representation" of it. Passing `-1` to `loader -> n` tells it to read all of the frames.

https://www.libvips.org/API/current/ctor.Image.gifload.html

This allows us to show the whole gif animation, but to also scale it down if needed.

Fixes: https://github.com/bullet-train-co/bullet_train/issues/2303

NOTE: We probably won't merge this PR here. I think we should hoist this file into the `core` repo and remove it from the starter repo. I'm just opening this PR to document something that I found that works.


jagthedrummer:
Closing in favor of https://github.com/bullet-train-co/bullet_train-core/pull/1273

PR #2454, opened 2026-04-16
Rails 8.1.3

jagthedrummer:


jagthedrummer:
The current test failures are caused by https://github.com/bullet-train-co/bullet_train-core/issues/1438 which is caused by https://github.com/rails/rails/pull/53281 which might be fixed by https://github.com/rails/rails/pull/56842 if we're lucky.

jagthedrummer:
> The current test failures are caused by [bullet-train-co/bullet_train-core#1438](https://github.com/bullet-train-co/bullet_train-core/issues/1438) which is caused by [rails/rails#53281](https://github.com/rails/rails/pull/53281) which might be fixed by [rails/rails#56842](https://github.com/rails/rails/pull/56842) if we're lucky.

For now I've added a monkey patch in `config/initializers/active_record_schema_dumper_monkey_patch.rb` to prevent the auto-sorting of columns in `schema.rb`.

Hopefully the Rails team will see fit to revert the auto-sorting change, or at least make it configurable so we don't have to maintain this file indefinitely.

The one remaining failing test is expected. We had to make a change in `core` to accommodate Rails 8.1 and the failing test is in the part of the CI pipeline that tests against the currently-published gems (which don't have the fix).

